PR #3, opened 2019-07-17
Revised README.md to be more sensible

mtompset:
Formatting of requirements section was ugly, and pip is no longer needed as of 1.24.0 it seems.


mtompset:
This incorporates the 'make' in the last one, by tweaking the instructions

PR #28, opened 2019-08-21
rewrite of authentication

Schparky:
Authentication is now structured for multi-org, multi-protocol SSO authentication, with SAML being the first protocol implementation.


application/models/organization_test.go
```diff
@@ -197,3 +184,80 @@ func TestValidateOrganization(t *testing.T) {
...
+				if err != nil {
+					t.Errorf("OrganizationFindByDomain() returned an error: %v", err)
+				} else if got.Uuid != test.want.Uuid {
+					t.Errorf("found %v, expected %v", got, test.want)
+				}
+			}
+		})
+	}
+
+	// delete org fixture and org domain by cascading delete
```
Baggerone: There is a `resetTables` functions in testutils.go already, do you want to use one of those here?

application/models/user_organization.go
```diff
@@ -56,3 +65,36 @@ func (u *UserOrganization) ValidateCreate(tx *pop.Connection) (*validate.Errors,
...
+	errs := validate.Validate(&validators.EmailIsPresent{Field: authEmail})
+	if len(errs.Errors) > 0 {
+		return []UserOrganization{}, fmt.Errorf("email address provided (%s) is not valid", authEmail)
+	}
+
+	where := "auth_email = ?"
+	params := []interface{}{authEmail}
+
+	if orgID != 0 {
+		where = "auth_email = ? AND organization_id = ?"
```
Baggerone:
```suggestion
		where += " AND organization_id = ?"
```

application/actions/auth.go
```diff
@@ -1,131 +1,199 @@
...
+		return authError(c, http.StatusInternalServerError, "UnableToLoadAuthProvider", "unable to load auth provider for organization")
 	}
 
-	authURL, err := sp.BuildAuthURL("")
-
-	return c.Redirect(302, authURL)
-}
-
-func AuthCallback(c buffalo.Context) error {
+	authResp := sp.Login(c)
```
Baggerone: I'm not sure what the reason for combining the two original Auth functions was.

In any case, could we move the rest of this function's code to another function that is named to clarify that it only gets called after authn has occurred?

application/models/organization_test.go
```diff
@@ -2,26 +2,13 @@ package models
...
-		Name:       "ACME",
-		Uuid:       orgUuid1,
-		AuthType:   "saml2",
-		AuthConfig: "[]",
-	}
-	if err := DB.Create(&org); err != nil {
-		t.Errorf("could not run test ... %v", err)
-		t.FailNow()
-	}
+	org, _ := createOrgFixtures(t)
```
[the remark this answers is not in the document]
Schparky: I tried to keep tests atomic/independent. If you see one that's not, I'll try to fix it.

application/models/organization_test.go
Schparky: I did this before @fillup created `resetTables()`. This is a bit simpler in that it doesn't need to clear out the `users` table.

application/actions/auth.go
Schparky: I like this suggestion. There is certainly room for cleanup here. I'd like to finish out this feature branch and start fresh with a new branch for refactoring.

application/models/organization_test.go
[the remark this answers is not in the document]
Baggerone: Personally, I'm OK with it, but there are two tests that call `createOrgFixtures()`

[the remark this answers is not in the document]
Schparky: Oh, I see. I was thinking differently. My main concern was the _content_ of the database: making certain that it's cleaned up after each test so that you don't key unique constraint violations and such. This is another concern, but definitely a tradeoff with code duplication.

PR #30, opened 2019-08-23
user info for posting

Schparky:
Moved /user call into gql by returning the authenticated user if the id is omitted from the gql user query


application/models/user.go
```diff
@@ -103,6 +106,9 @@ func (u *User) CreateAccessToken(org Organization, clientID string) (string, int
 }
 
 func (u *User) GetOrgIDs() []interface{} {
```
fillup: What is the reason/benefit of returning an array of interfaces rather than a concrete type?

application/models/user_test.go
```diff
@@ -413,60 +433,35 @@ func createUserFixtures(t *testing.T) (Organizations, User, UserOrganizations) {
...
+			for i, o := range got {
+				orgNames[i] = o.Name
+			}
+			if !reflect.DeepEqual(orgNames, test.want) {
+				t.Errorf("GetOrgIDs() = \"%v\", want \"%v\"", got, test.want)
+			}
+		})
 	}
-
+	resetTables(t) // Pack it in, Pack it out a/k/a "Leave No Trace"
```
fillup: Depending on all other tests to clean up before the next one runs sounds risky. Why not just clean up at the beginning of each test to ensure the state before interacting with it?

application/models/user.go
Schparky: I believe it's for the convenience of calling it from a `Where()`, like:

application/models/user_test.go
Schparky: Not sure what you mean by risky, but my intent was to just be as tidy as possible so that there is less risk of tests failing for no good reason.

PR #39, opened 2019-09-04
include posts with user

Schparky:


application/models/user.go
```diff
@@ -278,3 +287,29 @@ func (u *User) FindUserOrganization(org Organization) (UserOrganization, error)
 
 	return userOrg, nil
 }
+
+func (u *User) GetPosts(postRole string) ([]*Post, error) {
+	var postPtrs []*Post
+	if err := DB.Load(u, "PostsCreated", "PostsReceiving", "PostsProviding"); err != nil {
```
fillup: Since we're only returning one type of posts should we only load that one type and return it rather than load them all and return only a portion? It seems that'd save some load and time from the database.

PR #47, opened 2019-09-18
send new user to welcome page and refactor smaller functions out of AuthLogin

Baggerone:


Baggerone:
Billy,
Want to check the test's expected results to see if they are what you would want?

PR #51, opened 2019-09-24
Refactor auth to use separate functions for login and callback

Baggerone:


application/auth/types.go
```diff
@@ -10,8 +10,11 @@ const TypeSaml = "saml"
 // It is expected Login can be called multiple times, whether during initializing a login request or when
```
Schparky: Comment is outdated

application/actions/auth.go
```diff
@@ -235,51 +180,111 @@ func createAuthUser(
...
+
+	c.Session().Set(ClientIDSessionKey, clientID)
+
+	authEmail := c.Param(AuthEmailParam)
+	if authEmail == "" {
+		return authError(c, http.StatusBadRequest, "MissingAuthEmail", AuthEmailParam+" is required to login")
 	}
+	c.Session().Set(AuthEmailSessionKey, authEmail)
+
+	getOrSetReturnTo(c)
```
Schparky: Why is this handled differently than client ID and auth email? I think it would read easier to directly set the returnTo session key and rename `getOrSetReturnTo` to `getReturnTo`. Unless I'm missing something.

Baggerone: The main reason I left it unchanged was because it never returns an error - it's more flexible by just returning "/#" if it doesn't find anything.

PR #52, opened 2019-09-26
Feature/google auth

Baggerone:


application/auth/google/endpoint_legacy.go
```diff
@@ -0,0 +1,13 @@
+// +build !go1.9
+
+package google
+
+import (
+	"golang.org/x/oauth2"
+)
+
+// Endpoint is Google's OAuth 2.0 endpoint.
+var Endpoint = oauth2.Endpoint{
```
Schparky: How does this not conflict with the same variable in `endpoint.go`?

application/models/organization.go
```diff
@@ -55,10 +62,19 @@ func (o *Organization) ValidateUpdate(tx *pop.Connection) (*validate.Errors, err
...
 
 func (o *Organization) GetAuthProvider() (auth.Provider, error) {
+
 	if o.AuthType == AuthTypeSaml {
 		return saml.New([]byte(o.AuthConfig))
 	}
 
+	if o.AuthType == AuthTypeGoogle {
+		gCfg := google.GoogleConfig{}
+		json.Unmarshal([]byte(o.AuthConfig), &gCfg)
```
Schparky: The SAML provider does the Unmarshal inside New. For consistency, should they be the same?

application/auth/google/endpoint_legacy.go
Baggerone: I just borrowed those files, but it has to do with the top line `// +build !go1.9`

Schparky: Oh, I see. Thanks.

application/models/organization.go
Baggerone: Good point.

.env.example
```diff
@@ -6,4 +6,11 @@ AWS_ACCESS_KEY_ID=
 AWS_SECRET_ACCESS_KEY=
 AWS_DEFAULT_REGION=us-east-1
 AWS_REGION=us-east-1
-ROLLBAR_TOKEN=
\ No newline at end of file
+
+# For Google OAuth
+GOOGLE_SESSION_SECRET=abc123
+
+AUTH_CALLBACK_URL=https://wecarry.org/auth/callback
```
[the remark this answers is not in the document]
Baggerone: done

PR #53, opened 2019-09-27
implemented files model

Schparky:
* `posts` has a one-to-many relationship with `files`, via a join table with a unique `file_id` foreign key.
* `posts` also has a foreign key to `files` to represent the primary post photo
* `users` has a foreign key to `files` to represent the profile photo

No API implementation has been done yet, but expect a RESTful `POST` endpoint to upload a file, and GraphQL queries and mutations to query and link files at appropriate places in the data model.


application/aws/aws.go
```diff
@@ -0,0 +1,154 @@
...
+	var a awsConfig
+	a.awsAccessKeyID = envy.Get(AwsS3AccessKeyIDEnv, "")
+	a.awsSecretAccessKey = envy.Get(AwsS3SecretAccessKeyEnv, "")
+	a.awsEndpoint = envy.Get(AwsS3EndpointEnv, "")
+	a.awsRegion = envy.Get(AwsS3RegionEnv, "")
+	a.awsS3Bucket = envy.Get(AwsS3BucketEnv, "")
+
+	if disableSSL, err := strconv.ParseBool(envy.Get(AwsS3DisableSSLEnv, "false")); err == nil {
+		a.awsDisableSSL = disableSSL
+	} else {
```
Baggerone: The `else` block is redundant, right, i.e. the initial value of a bool is already false? Or, do you maybe want to deal with the error?

application/aws/aws.go
```diff
@@ -0,0 +1,154 @@
...
+		a.awsDisableSSL = false
+	}
+
+	if len(a.awsEndpoint) > 0 {
+		// a non-empty endpoint means minIO is in use, which doesn't support the S3 object URL scheme
+		a.getPresignedUrl = true
+	} else {
```
Baggerone: Redundant `else` block?

application/aws/aws.go
```diff
@@ -0,0 +1,154 @@
...
+		S3ForcePathStyle: aws.Bool(len(config.awsEndpoint) > 0),
+	})
+	svc := s3.New(sess)
+
+	return svc, err
+}
+
+func getObjectURL(config awsConfig, svc *s3.S3, key string) (ObjectUrl, error) {
+	var objectUrl ObjectUrl
+	if config.getPresignedUrl {
```
Baggerone: Just stylistic ...
  You could use the lines from the `else` block first and not have a long, nested `if` block ...

   ```
  if !config.getPresignedUrl {
      objectUrl.Url = fmt.Sprintf("https://%s.s3.amazonaws.com/%s", config.awsS3Bucket, url.PathEscape(key))
      objectUrl.Expiration = time.Date(9999, time.December, 31, 0, 0, 0, 0, time.UTC)
      return objectUrl, nil
  }
```

File: application/aws/aws.go [hunk 4]
```diff
@@ -0,0 +1,154 @@
...
+		req, _ := svc.GetObjectRequest(&s3.GetObjectInput{
+			Bucket: aws.String(config.awsS3Bucket),
+			Key:    aws.String(key),
+		})
+
+		if newUrl, err := req.Presign(urlLifespan); err == nil {
+			objectUrl.Url = newUrl
+			// return a time slightly before the actual url expiration to account for delays
+			objectUrl.Expiration = time.Now().Add(urlLifespan - time.Minute)
+		} else {
```
Baggerone: No `else` needed here, right?

application/aws/aws.go
```diff
@@ -0,0 +1,154 @@
...
+// credentials to access. It may reference a file with public_read access or it may be a pre-signed URL.
+func GetFileURL(key string) (ObjectUrl, error) {
+	config := GetS3ConfigFromEnv()
+
+	svc, err := CreateS3Service(config)
+	if err != nil {
+		return ObjectUrl{}, err
+	}
+
+	objectUrl, err := getObjectURL(config, svc, key)
```
Baggerone: Here, you could just `return getObjectURL(config, svc, key)`, right?

application/models/file.go
```diff
@@ -0,0 +1,152 @@
...
+	allowedTypes := []string{
+		"image/bmp",
+		"image/gif",
+		"image/jpeg",
+		"image/png",
+		"image/webp",
+	}
+
+	detectedType := http.DetectContentType(content)
+	for _, t := range allowedTypes {
```
Baggerone: or ...
```
    if domain.IsStringInSlice(detectedType, allowedTypes) {
        return detectedType, nil
    }
```

application/models/file_test.go
```diff
@@ -0,0 +1,267 @@
...
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			vErr, _ := test.file.Validate(DB)
+			if test.wantErr {
+				if vErr.Count() == 0 {
+					t.Errorf("Expected an error, but did not get one")
+				} else if len(vErr.Get(test.errField)) == 0 {
+					t.Errorf("Expected an error on field %v, but got none (errors: %v)", test.errField, vErr.Errors)
+				}
+			} else if (test.wantErr == false) && (vErr.HasAny()) {
```
Baggerone:
```suggestion
			} else if (vErr.HasAny() {
```

docker-compose.yml
```diff
@@ -18,6 +18,9 @@ services:
       UI_URL: http://wecarry.local:5000
       DATABASE_URL: postgres://wecarry:wecarry@db:5432/wecarry?sslmode=disable
       GO_ENV: development
+      AWS_S3_ENDPOINT: http://minio:9000
+      AWS_S3_DISABLE_SSL: "true"
+      AWS_S3_BUCKET: gtis-hca-dev
```
Baggerone: Should all of these `gtis-hca-dev` values be changed to substitute something less identifiable?

application/aws/aws.go
Schparky: It's needed in order to return the error from `Presign()`

application/aws/aws.go
```diff
@@ -0,0 +1,148 @@
...
+
+	if newUrl, err := req.Presign(urlLifespan); err == nil {
+		objectUrl.Url = newUrl
+		// return a time slightly before the actual url expiration to account for delays
+		objectUrl.Expiration = time.Now().Add(urlLifespan - time.Minute)
+	} else {
+		return objectUrl, err
+	}
+
+	return objectUrl, nil
```
Baggerone: If you want to leave this as it is to avoid questions about the scope of `err`, that's fine.

But just to clarify my original comment ...
Instead of having the `else` block, I think you can just `return objectUrl, err`.

If `err` wasn't `nil`, then this returns the error.  If `err` was `nil`, it returns `nil`. Either way, the right thing gets returned.

Schparky: I see. I think I prefer it this way, with err being scoped to the if block. Thanks for clarifying.

PR #54, opened 2019-09-30
fix auth error to redirect to UI without details

Baggerone:


application/actions/auth.go
```diff
@@ -311,46 +323,44 @@ func AuthCallback(c buffalo.Context) error {
...
 }
 
-// returnAuthError takes a error code and message and renders AuthResponse to json and returns
-func authError(c buffalo.Context, status int, code, message string) error {
-	resp := AuthResponse{
-		Error: &AuthError{
-			Code:    code,
-			Message: message,
-		},
+func authError(c buffalo.Context, code, message string, extras map[string]interface{}) error {
```
Schparky: Is it possible/helpful to define `extras` as a variadic so `emptyExtras` isn't needed?

application/actions/auth.go
```diff
@@ -42,6 +42,11 @@ const ReturnToSessionKey = "ReturnTo"
 // http param for token type
 const TokenTypeParam = "token-type"
 
+// environment variable key for the UI's URL
+const UIURLEnv = "UI_URL"
```
Schparky: This could be used in app.og and auth_test.go also.

application/actions/auth.go
Baggerone: Good point.

application/actions/auth.go
Baggerone: Done

application/actions/auth.go
Baggerone: Ha, ha.  Making the change showed me I had a bug in authError ... I was merging `extras` into `allExtras`, but then not using `allExtras` anymore.

application/actions/auth.go
```diff
@@ -311,46 +320,47 @@ func AuthCallback(c buffalo.Context) error {
...
+		for k, v := range e {
+			allExtras[k] = v
+		}
 	}
 
-	return c.Render(status, render.JSON(resp))
+	domain.Error(c, message, allExtras)
+
+	uiUrl := envy.Get(UIURLEnv, "") + "/#/login?error=true"
+	return c.Redirect(http.StatusFound, uiUrl)
```
fillup: That isn't a redirect status code, should be temporary redirect, 302. Also `error=true` is too ambiguous to do anything intelligent with. Why aren't we returning more info in the json response for auth requests and redirecting on callback requests?

Baggerone: I was just trying to implement what you had suggested on slack ...
```Phillip S:spiral_calendar_pad:  5 days ago
The redirect code would be 302 as a temporary redirect, and the URL would probably need to be something like {UI_URL}/#/login?error=true```

302 is the value of http.StatusFound.

If I remember right, Billy was thinking that should be a 500.

fillup: Hmm, `StatusFound` really a 302?

Also the redirect is for the callback endpoint where it is not an api call from the UI but a redirection from auth provider

fillup: Hmm, `StatusFound` really a 302?

Also the redirect is for the callback endpoint where it is not an api call from the UI but a redirection from auth provider

fillup: I think in the context of an auth callback error just redirecting with a generic `error=ture` could be sufficient, but during authrequest, since it is already an api call just returning json would be best to handle the multiple possible scenarios (a URL to redirect to to login, a list of org to choose from, or an error). By returning the error here we can handle different reasons more specifically, like if we don't know how to login for a given domain name, present a message about how to get in touch to sign up, etc.

PR #55, opened 2019-10-02
User and Post photos

Schparky:
* new `photoURL` property on `User`
* new `photo` and `files` properties on `Post`
* new `updateUser` mutation for attaching a new profile photo
* new `photoID` property on input to `UpdatePost` mutation

Note that there is not yet a mutation for adding files, other than the primary image, to posts.


application/actions/upload.go
```diff
@@ -0,0 +1,59 @@
...
+	URL         string           `json:"url,omitempty"`
+	ContentType string           `json:"content_type,omitempty"`
+	Size        int              `json:"size,omitempty"`
+}
+
+// UploadHandler responds to POST requests at /upload
+func UploadHandler(c buffalo.Context) error {
+	f, err := c.File(FileFieldName)
+	if err != nil {
+		return c.Render(http.StatusInternalServerError, render.JSON(UploadResponse{
```
Baggerone: Maybe we should include a TODO comment about logging these errors by calling a function that calls domain.Error(), similar to actions.authRequestError()

application/actions/upload_test.go
```diff
@@ -0,0 +1,129 @@
...
+	"github.com/silinternational/wecarry-api/models"
+)
+
+// UploadFixtures is for returning fixtures from Fixtures_Upload
+type UploadFixtures struct {
+	ClientID    string
+	AccessToken string
+}
+
+// Fixtures_Upload creates fixtures for the Test_Upload test
```
Baggerone: Do you prefer having the fixture in the same file like this?  It seems neater to me to have it in a different file, so instead of scrolling up and down, you're just switching tabs.  Also, it means the test file stays cleaner.

Your choice.

application/aws/aws.go
```diff
@@ -146,3 +143,28 @@ func GetFileURL(key string) (ObjectUrl, error) {
...
+// exists, it will not return an error.
+func CreateS3Bucket() error {
+	config := GetS3ConfigFromEnv()
+
+	svc, err := CreateS3Service(config)
+	if err != nil {
+		return err
+	}
+
+	if _, err := svc.CreateBucket(&s3.CreateBucketInput{
```
Baggerone: Maybe have the CreateBucketInput done in the line above, to improve readability?

fillup: The bucket should be created by Terraform along with the user and appropriate policy to restrict user access and then fed into the environment

Schparky: Good point. Nowhere in the app is this called. I'll try moving it back to test code, but I've had trouble in the past accessing test code from another package. I guess I could just create two copies of this code.

application/actions/upload_test.go
Schparky: Yes, I do, for organizational clarity. I do appreciate your concern. Would an editor split view work for this?

PR #56, opened 2019-10-02
rename GraphQL identifiers

Schparky:
* use standard naming for inputs: `<verb><object>Input`
* shorten `OrganizationDomain` to `Domain`

### DIFF
--- a/application/gqlgen/mutations.go
+++ b/application/gqlgen/mutations.go
@@ -12,9 +12,9 @@ import (
 
 type mutationResolver struct{ *Resolver }
 
-func (r *mutationResolver) CreateMessage(ctx context.Context, input NewMessage) (*models.Message, error) {
+func (r *mutationResolver) CreateMessage(ctx context.Context, input CreateMessageInput) (*models.Message, error) {
 	cUser := models.GetCurrentUserFromGqlContext(ctx, TestUser)
-	message, err := ConvertGqlNewMessageToDBMessage(input, cUser)
+	message, err := ConvertGqlCreateMessageInputToDBMessage(input, cUser)
 	if err != nil {
 		domain.Error(models.GetBuffaloContextFromGqlContext(ctx), err.Error(), domain.NoExtras)
 		return &models.Message{}, err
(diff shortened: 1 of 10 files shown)


application/gqlgen/mutations.go
```diff
@@ -70,7 +70,7 @@ func (r *mutationResolver) UpdateOrganization(ctx context.Context, input Updated
 	return &org, err
 }
 
-func (r *mutationResolver) CreateOrganizationDomain(ctx context.Context, input NewOrganizationDomain) ([]*models.OrganizationDomain, error) {
+func (r *mutationResolver) CreateDomain(ctx context.Context, input CreateDomainInput) ([]*models.OrganizationDomain, error) {
```
Baggerone: Since the model is called `OrganizationDomain`, can this at least be `CreateOrgDomain`?
The same would be true for the other similar changes you made removing `Organization`. 

`Domain` is a very generic word and it seems better to specify what we mean here just to avoid confusion or a potential additional use of `domain` related to other objects, etc.

Schparky: I'm open to it. Let me elaborate on the thoughts I had while making these changes:
* The name of the model doesn't have any bearing on good API decisions. (I'd like to change Post to Request and Thread to Conversation also.) If we want consistency, it's better to change the model name.
* "Organization Domain" implies that multiple orgs can have the same domain. In other words, the combination of Organization and Domain is unique, but not Domain by itself. In my opinion, in making the name simply Domain, it becomes more clear that it stands on its own in that sense.
* If a future need for a different definition of "domain" arises, a clarifying adjective or synonym can be used.

Baggerone: @fillup What do you think about this?  (In gql ... "OrganizationDomain" vs. "Domain" vs "OrgDomain")

fillup: If we can't think of another kind of Domain for future use then I'm okay with dropping Organization in this case, but I like that OrganizationDomain ties organizations and domains together, since domains do not exist on their own in how we use them. 

As for Post vs Request. we chose Post because eventually Posts will be Requests, or Offers, or maybe other things, so we didn't want to constrict the use or set ourselves up for a confusing/contradicting name. 

As far as I can think through Thread vs Conversation I don't have an argument either way, though I don't see benefit one way or the other either so I personally would not rename it but don't really object either.

PR #57, opened 2019-10-03
add a dev tool to set up the development environment

Schparky:
At present, the only thing this does is to create a bucket in minIO.


dev/dev.go
```diff
@@ -0,0 +1,96 @@
...
+	"strconv"
+
+	"github.com/aws/aws-sdk-go/aws"
+
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/s3"
+)
+
```
Baggerone: In dev.go, can you pull things in from application/aws by adding this to your imports ...
`wecarryaws "github.com/silinternational/wecarry-api/aws"`

Then in the dev folder, you would need to do `go get ./...`

And then in the go.mod file, you would need to add ...

`replace github.com/silinternational/wecarry-api v0.0.0 => ../application`

Then run `go get ./...` again

Schparky: Great suggestion. Works like a charm, with one minor side-effect. Goland complains at startup because its relative path for `wecarry-api` should be `../application`. I still went with your suggestion, hoping to find a solution for this annoyance.

PR #59, opened 2019-10-03
limit the list of threads on a post to only those for the current user

Schparky:


application/models/post.go
```diff
@@ -141,11 +142,24 @@ func (p *Post) GetOrganization(fields []string) (*Organization, error) {
...
 
-func (p *Post) GetThreads(fields []string) ([]*Thread, error) {
+// GetThreads finds all threads on this post in which the given user is participating
+func (p *Post) GetThreads(fields []string, user User) ([]*Thread, error) {
+	if err := DB.Load(p, "Threads"); err != nil {
+		return nil, fmt.Errorf("error getting threads for post id %v ... %v", p.ID, err)
+	}
+
 	var threads []*Thread
+	for i, t := range p.Threads {
```
Baggerone: Would it be better to do this with a LeftJoin, similar to gqlgen/thread.go->MyThreads?

Schparky: Possibly, but I tried it and it just seemed messier. I'd like to do a test and see what the performance difference is. Maybe another day, haha.

PR #65, opened 2019-10-04
Feature/fix logout

Baggerone:
As it was, we weren't getting the hydrated organization from the access token in order to know what the auth provider should be.


application/domain/domain.go
```diff
@@ -26,6 +26,9 @@ const (
 	MaxFileSize              = 1 << 20 // 1 Mebibyte
 )
```
Schparky: Want to include `UIURLEnv` inside the `const` block?

PR #66, opened 2019-10-04
tweak GetOrg and make a real test

Baggerone:


application/models/user_access_token.go
```diff
@@ -79,26 +79,30 @@ func (u *UserAccessToken) FindByBearerToken(bearerToken string) error {
 }
 
 // GetOrganization returns the Organization of the UserOrganization of the UserAccessToken
-//  I'm not using DB.Eager() or DB.Load() because they seemed to keep hanging.
-//  I wonder if it was because of the has_many  <--->  belongs_to relations
 func (u *UserAccessToken) GetOrganization() (Organization, error) {
 	if u.UserOrganizationID <= 0 {
 		return Organization{}, fmt.Errorf("user access token id %v has no user organization", u.ID)
```
Schparky: Is this check necessary? (same a few lines down)

application/models/user_access_token_test.go
```diff
@@ -145,36 +147,138 @@ func CreateUserAccessTokenFixtures(t *testing.T, user User, userOrgs UserOrganiz
...
+		},
+	}
+	for i := range userOrgs {
+		if err := ms.DB.Create(&userOrgs[i]); err != nil {
+			t.Errorf("could not create test user org ... %v. uo = %+v", err, userOrgs[i])
+			t.FailNow()
+		}
+	}
+
+	return orgs, users, userOrgs
```
Schparky: After I saw fillup use structs to return fixtures, I decided to use that idea going forward. If you prefer this, that's fine with me.

application/models/user_access_token_test.go
```diff
@@ -145,36 +147,138 @@ func CreateUserAccessTokenFixtures(t *testing.T, user User, userOrgs UserOrganiz
...
+	ResetTables(t, ms.DB)
+
+	unique := domain.GetUuid().String()
+
+	// Load Organization test fixtures
+	orgs := []Organization{
+		{
+			Name:       fmt.Sprintf("ACME-%s", unique),
+			Uuid:       domain.GetUuid(),
+			AuthType:   "saml2",
```
Schparky:
```suggestion
			AuthType:   AuthTypeSaml,
```

application/models/user_access_token.go
Baggerone: OK. I was worried about panics, but it looks like your right about Load catching it.

PR #67, opened 2019-10-07
Renew access token when accessed

Schparky:
Renew user access token when it is accessed and change default expiration to 1 hour (from 8 hours)

### DIFF
--- a/application/models/user.go
+++ b/application/models/user.go
@@ -251,7 +251,7 @@ func (u *User) FindByAccessToken(accessToken string) error {
 	}
 
 	*u = userAccessToken.User
-	return nil
+	return userAccessToken.Renew()
 }
 
 func (u *User) FindByUUID(uuid string) error {
--- a/application/models/user.go
+++ b/application/models/user.go
@@ -5,6 +5,7 @@ import (
 	"crypto/sha256"
 	"encoding/json"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/pkg/errors"
--- a/application/models/user.go
+++ b/application/models/user.go
@@ -5,8 +5,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"log"
-	"math/rand"
-	"strconv"
 	"time"
 
 	"github.com/pkg/errors"
@@ -14,7 +12,6 @@ import (
 
 	"github.com/silinternational/wecarry-api/auth"
 
-	"github.com/gobuffalo/envy"
 	"github.com/gobuffalo/nulls"
 	"github.com/gobuffalo/pop"
 	"github.com/gobuffalo/validate"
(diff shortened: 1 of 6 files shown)


application/models/user.go
```diff
@@ -254,7 +251,7 @@ func (u *User) FindByAccessToken(accessToken string) error {
 	}
 
 	*u = userAccessToken.User
-	return nil
+	return userAccessToken.Renew()
```
Baggerone: It seems like this should be a separate call, partly so we don't get confused if there's an error and partly because I  wouldn't expect it to have much to do with finding the access token.

application/models/user_access_token.go
```diff
@@ -102,3 +106,39 @@ func (u *UserAccessToken) GetOrganization() (Organization, error) {
 
 	return uOrg.Organization, nil
 }
+
+func createAccessTokenExpiry() time.Time {
+	lifetime := envy.Get("ACCESS_TOKEN_LIFETIME", "3600")
```
Baggerone: Want to make the default for this and `lifetimeSeconds` constants that match?

application/models/user_access_token.go
```diff
@@ -102,3 +106,39 @@ func (u *UserAccessToken) GetOrganization() (Organization, error) {
...
+
+	accessToken := string(b)
+
+	return accessToken
+}
+
+func (uat *UserAccessToken) Renew() error {
+	uat.ExpiresAt = createAccessTokenExpiry()
+	if err := DB.Update(uat); err != nil {
+		return fmt.Errorf("error renewing access token, %s", err)
```
Baggerone: Somewhere in the chain of calls, this error should get logged.

application/models/user.go
Schparky: Agreed. I refactored to make this possible.

application/models/user_access_token.go
Schparky: Good call. Added a const to domain.go.

application/models/user_access_token.go
Schparky: Done.

application/models/user_access_token.go
```diff
@@ -102,3 +109,64 @@ func (u *UserAccessToken) GetOrganization() (Organization, error) {
 
 	return uOrg.Organization, nil
 }
+
+func createAccessTokenExpiry() time.Time {
+	envLifetime := envy.Get("ACCESS_TOKEN_LIFETIME", strconv.Itoa(domain.AccessTokenLifetimeSeconds))
+
+	lifetimeSeconds, err := strconv.Atoi(envLifetime)
+	if err != nil {
+		lifetimeSeconds = domain.AccessTokenLifetimeSeconds
```
Baggerone:
```suggestion
        // TODO Ensure this gets logged so that we know our env var is bad
		lifetimeSeconds = domain.AccessTokenLifetimeSeconds
```

PR #71, opened 2019-10-08
event listeners

Baggerone:
Is this a good way to structure things?
(Sorry, I couldn't resist tweaking a few things that weren't really related)


application/models/user_access_token.go
```diff
@@ -159,15 +158,37 @@ func (u *UserAccessToken) GetUser() (User, error) {
...
+			return true, fmt.Errorf("unable to delete expired userAccessToken, id: %v", u.ID)
 		}
-		return true
+		return true, nil
 	}
-	return false
+	return false, nil
+}
+
+func UserAccessTokensDeleteExpired() (int, error) {
```
Schparky:
```suggestion
func (u *UserAccessTokens) DeleteExpired() (int, error) {
```

application/models/user_access_token_test.go
```diff
@@ -520,7 +524,95 @@ func (ms *ModelSuite) TestUserAccessToken_Renew() {
...
+	}
+
+	gotToken := uats[0].AccessToken
+	wantToken := f.UserAccessTokens[1].AccessToken
+
+	if gotToken != wantToken {
+		t.Errorf("Wrong token remaining. \nWant %v \n Got %v", wantToken, gotToken)
+	}
+
+}
```
Schparky: suggest using `ms.Equal` for these comparisons

application/models/user_access_token_test.go
```diff
@@ -520,7 +524,95 @@ func (ms *ModelSuite) TestUserAccessToken_Renew() {
...
+	ResetTables(t, ms.DB)
+
+	f := createFixtures_UserAccessTokensDeleteExpired(ms, t)
+
+	got, err := UserAccessTokensDeleteExpired()
+
+	if err != nil {
+		t.Errorf("Unexpected error: %v", err)
+		return
+	}
```
Schparky: `ms.NoError(err)` is a concise way to do this

application/actions/auth.go
```diff
@@ -406,10 +406,20 @@ func SetCurrentUser(next buffalo.Handler) buffalo.Handler {
 
 		var userAccessToken models.UserAccessToken
 		err := userAccessToken.FindByBearerToken(bearerToken)
-		if err != nil || userAccessToken.IsExpired() {
+		if err != nil {
+			domain.ErrLogger.Print(err.Error())
 			return c.Error(401, fmt.Errorf("invalid bearer token"))
 		}
 
+		isExpired, err := userAccessToken.DeleteIfExpired()
```
Schparky: I like the function name change.

application/domain/domain.go
```diff
@@ -24,14 +26,30 @@ const (
 	EmptyUUID                  = "00000000-0000-0000-0000-000000000000"
 	DateFormat                 = "2006-01-02"
 	MaxFileSize                = 1 << 20 // 1 Mebibyte
-	UIURLEnv                   = "UI_URL"
 	AccessTokenLifetimeSeconds = 3600
+	DateTimeFormat             = "2006-01-02 15:04:05"
+
+	// Environment Variables
```
Schparky: Good groupings. I've seen this done with new `const` blocks. Thoughts on that?

application/listeners/listeners.go
```diff
@@ -0,0 +1,71 @@
+package listeners
+
+import (
+	"github.com/gobuffalo/events"
+	"github.com/silinternational/wecarry-api/domain"
+	"github.com/silinternational/wecarry-api/models"
+	"time"
+)
+
+// For new listener functions, register them at the end of the
```
Schparky: Why not just put the slice here?

Baggerone: Ha, ha.  That's my Python experience showing through.

application/models/user_access_token_test.go
Baggerone: Done.

application/models/user_access_token_test.go
Baggerone: Good, thanks.

application/models/user_access_token.go
Baggerone: Done.

Baggerone:
I think it's preferable leaving the Loggers as globals rather than in a context that gets passed to all the functions.  I see them as more like constants.  Once we set their output in the init(), we're only using their methods rather than modifying their attributes.

application/domain/domain.go
Baggerone: Sounds good.

PR #76, opened 2019-10-11
crypto/rand instead of math/rand

Baggerone:
math/rand is predictable and not really suitable for access tokens

### DIFF
--- a/application/models/models.go
+++ b/application/models/models.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"crypto/rand"
 	"database/sql"
+	"encoding/base64"
 	"fmt"
 	"github.com/gobuffalo/events"
 	"github.com/gobuffalo/validate/validators"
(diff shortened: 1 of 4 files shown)


application/models/models.go
```diff
@@ -33,6 +36,11 @@ func init() {
 		log.Fatal(err)
 	}
 	pop.Debug = env == "development"
+
+	// Just make sure we can use the crypto/rand library on our system
+	if _, err = rand.Read(RandomBytes); err != nil {
+		log.Fatal(fmt.Errorf("error using crypto/rand ... %v", err))
+	}
```
Schparky: I'm not clear on how often `init()` is executed, but isn't it on every API request? If so, why do this every time? `RandomBytes` wouldn't have to be a global if this wasn't here.

Baggerone: It only gets run the first time the package is imported.  I was thinking that if it was going to get an error ever, it would get it every time. So, why not catch the error at the start and then not worry about checking for errors later.

PR #77, opened 2019-10-11
Ensure that a user created from an auth user gets unique nickname

Baggerone:
I considered getting a semi-random prefix for the nickname, but figured it wasn't worth the added complexity and bug-risk just for a few users.

### DIFF
--- a/application/models/user.go
+++ b/application/models/user.go
@@ -174,7 +174,7 @@ func (u *User) FindOrCreateFromAuthUser(orgID int, authUser *auth.User) error {
 		u.Uuid = domain.GetUuid()
 
 		u.Nickname = authUser.Nickname
-		if err := u.getUniqueNickname(); err != nil {
+		if err := u.uniquifyNickname(); err != nil {
 			return err
 		}
 	}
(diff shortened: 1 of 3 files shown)


application/models/user.go
```diff
@@ -354,3 +350,38 @@ func (u *User) Save() error {
...
+
+func (u *User) getUniqueNickname() error {
+
+	simpleNN := u.Nickname
+	if simpleNN == "" {
+		last := ""
+		if len(u.LastName) > 0 {
+			last = u.LastName[:1]
+		}
+		simpleNN = u.FirstName + last
```
Schparky: Can this be reduced to:
```go
if len(u.LastName) > 0 {
    simpleNN = u.FirstName + u.LastName[:1]
}
```

application/models/user.go
```diff
@@ -354,3 +350,38 @@ func (u *User) Save() error {
 
 	return DB.Save(u)
 }
+
+func (u *User) getUniqueNickname() error {
```
Schparky: Since the unique nickname isn't a return value, should the name be something like `uniquifyNickname`? Since it doesn't save to the DB, I might be more inclined to keep the name and return the new nickname for the caller to handle. Since it's a non-exported function, either way is probably fine.

application/models/user_test.go
```diff
@@ -696,3 +696,74 @@ func (ms *ModelSuite) TestUser_AttachPhoto() {
 		ms.Fail("user.GetPhotoURL failed, %s", err)
 	}
 }
+
+func CreateUserFixturesForNicknames(ms *ModelSuite, t *testing.T) User {
+	ResetTables(t, ms.DB)
+
+	unique := domain.GetUuid().String()
```
Schparky: I like `t.Name()` because it's informative. Granted, it may not work if test interations > 1. Not sure about that.
```suggestion
	unique := t.Name()
```

application/models/user.go
Baggerone: Done.

application/models/user.go
Baggerone: Done.

application/models/user_test.go
Baggerone: Done.

PR #78, opened 2019-10-11
Email notifications for new messages

Schparky:
Implemented the structure for notifications, and implemented "new message" notification by email.


application/models/message.go
```diff
@@ -83,3 +88,43 @@ func (m *Message) GetThread(requestFields []string) (*Thread, error) {
...
+	uiUrl := envy.Get(domain.UIURLEnv, "")
+	for _, tp := range m.Thread.Participants {
+		if tp.ID == m.SentBy.ID {
+			continue
+		}
+		e := events.Event{
+			Kind:    domain.EventApiMessageCreated,
+			Message: "New Message from " + m.SentBy.Nickname,
+			Payload: events.Payload{
+				"toName":         tp.Nickname,
```
Baggerone: Instead of all these string keys, how about doing something like this (assuming you put the struct outside the function.  I added a comment in the listeners about what to do there.)  ...
```
	type MessageCreatedEventData struct {
		MessageCreatorNickName string
		MessageCreatorEmail    string
		MessageContent         string
		PostUUID               string
		PostTitle              string
		ThreadUUID             string
		ThreadParticipants     []struct{ Nickname, Email string }
	}

	eventData := MessageCreatedEventData{
		MessageCreatorNickName: m.SentBy.Nickname,
		MessageCreatorEmail:    m.SentBy.Email,
		MessageContent:         m.Content,
		PostUUID:               m.Thread.Post.Uuid.String(),
		PostTitle:              m.Thread.Post.Title,
		ThreadUUID:             m.Thread.Uuid.String(),
	}

	for _, tp := range m.Thread.Participants {
		if tp.ID == m.SentBy.ID {
			continue
		}
		eventData.ThreadParticipants = append(eventData.ThreadParticipants,
			struct{ Nickname, Email string }{Nickname: tp.Nickname, Email: tp.Email})
	}

	e := events.Event{
		Kind:    domain.EventApiMessageCreated,
		Message: "New Message from " + m.SentBy.Nickname,
		Payload: events.Payload{
			"eventData": eventData,
		},
	}

	emitEvent(e)
```

application/notifications/email/dummy.go
```diff
@@ -0,0 +1,47 @@
...
+		subject: "new message",
+		body:    "You have a new message.",
+	},
+}
+
+func (t *DummyService) Send(msg Message) error {
+	template, ok := dummyTemplates[msg.TemplateName]
+	if !ok {
+		domain.ErrLogger.Printf("invalid template name: %s", msg.TemplateName)
+	}
```
Baggerone: Do you need a `return` here?

application/listeners/listeners.go
```diff
@@ -76,6 +85,44 @@ func userCreated(e events.Event) {
 	domain.Logger.Printf("%s User Created ... %s", domain.GetCurrentTime(), e.Message)
 }
 
+func sendNewMessageNotification(e events.Event) {
+	if e.Kind != domain.EventApiMessageCreated {
+		return
+	}
+	domain.Logger.Printf("%s Message Created ... %s", domain.GetCurrentTime(), e.Message)
+
```
Baggerone: If you make the change in the message.Create() function that I described, you'll need to do something like this here ...
```
  mEData := e.Payload["eventData"].(models.MessageCreatedEventData)

  data := map[string]interface{} {
    "postTitle": mEData.PostTitle,
    ...
```
  }

PR #85, opened 2019-10-18
skeleton for status transition events

Baggerone:
Before I work on tests and filling this out more, I wanted to make sure I'm on the right track.

### DIFF
--- a/application/domain/domain.go
+++ b/application/domain/domain.go
@@ -59,6 +59,7 @@ const (
 	MessageTemplateRequestFromReceivedToAccepted   = "request_from_received_to_accepted"
 	MessageTemplateRequestFromCommittedToDelivered = "request_from_committed_to_delivered"
 	MessageTemplateRequestFromAcceptedToDelivered  = "request_from_accepted_to_delivered"
+	MessageTemplateRequestFromReceivedToDelivered  = "request_from_received_to_delivered"
 	MessageTemplateRequestFromCompletedToDelivered = "request_from_completed_to_delivered"
 	MessageTemplateRequestFromAcceptedToReceived   = "request_from_accepted_to_received"
 	MessageTemplateRequestFromCompletedToReceived  = "request_from_completed_to_received"
--- a/application/domain/domain.go
+++ b/application/domain/domain.go
@@ -41,15 +41,32 @@ const (
 
 // Event Kinds
 const (
-	EventApiUserCreated      = "api:user:created"
-	EventApiAuthUserLoggedIn = "api:auth:user:loggedin"
-	EventApiMessageCreated   = "api:message:created"
+	EventApiUserCreated       = "api:user:created"
+	EventApiAuthUserLoggedIn  = "api:auth:user:loggedin"
+	EventApiMessageCreated    = "api:message:created"
+	EventApiPostStatusUpdated = "api:post:status:updated"
 )
 
 // Notification Message Template Names
 const (
-	MessageTemplateNewMessage = "new_message"
-	MessageTemplateNewRequest = "new_request"
+	MessageTemplateNewMessage                      = "new_message"
+	MessageTemplateNewRequest                      = "new_request"
+	MessageTemplateRequestFromCommittedToOpen      = "request_from_committed_to_open"
+	MessageTemplateRequestFromAcceptedToOpen       = "request_from_accepted_to_open"
+	MessageTemplateRequestFromOpenToCommitted      = "request_from_open_to_committed"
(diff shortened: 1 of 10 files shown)


application/listeners/listeners.go
```diff
@@ -128,3 +135,36 @@ func sendNewMessageNotification(e events.Event) {
...
+	if !ok {
+		domain.ErrLogger.Print("unable to parse Post Status Updated event payload")
+		return
+	}
+
+	if pEData.Post.Type != models.PostTypeRequest {
+		return
+	}
+
+	switch pEData.NewStatus {
```
Schparky: Instead of the two levels of `switch`es, would it work to have a map of templates where the key is the two statuses concatenated?

application/domain/domain.go
```diff
@@ -41,15 +41,32 @@ const (
...
+	MessageTemplateRequestFromCommittedToDelivered = "request_from_committed_to_delivered"
+	MessageTemplateRequestFromAcceptedToDelivered  = "request_from_accepted_to_delivered"
+	MessageTemplateRequestFromCompletedToDelivered = "request_from_completed_to_delivered"
+	MessageTemplateRequestFromAcceptedToReceived   = "request_from_accepted_to_received"
+	MessageTemplateRequestFromCompletedToReceived  = "request_from_completed_to_received"
+	MessageTemplateRequestFromDeliveredToCompleted = "request_from_delivered_to_completed"
+	MessageTemplateRequestFromReceivedToCompleted  = "request_from_received_to_completed"
+	MessageTemplateRequestFromOpenToRemoved        = "request_from_open_to_removed"
+	MessageTemplateRequestFromCommittedToRemoved   = "request_from_committed_to_removed"
+	MessageTemplateRequestFromAcceptedToRemoved    = "request_from_accepted_to_removed"
```
Schparky: Do the "reverted" status changes need a notification? And "*_to_removed"?

application/listeners/posthelpers.go
```diff
@@ -0,0 +1,129 @@
...
+	"github.com/silinternational/wecarry-api/models"
+	"github.com/silinternational/wecarry-api/notifications"
+)
+
+type msgRecipient struct {
+	nickname string
+	email    string
+}
+
+func getRecipients(eData models.PostStatusEventData) []msgRecipient {
```
Schparky: Is there a way to make this usable by other event listeners?

application/models/post.go
```diff
@@ -4,9 +4,10 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"github.com/gobuffalo/buffalo"
+	"github.com/gobuffalo/events"
 	"time"
 
-	"github.com/gobuffalo/buffalo"
 	"github.com/gobuffalo/nulls"
```
Schparky: Are you manually adding imports? `goimports` sorts differently than this.

application/domain/domain.go
Baggerone: For sure in some cases.  For example a provider commits and then reverts, we would need to tell the requestor, right?
Or if a requestor removes a post that has a provider, we would need to tell the provider.

application/models/post.go
Baggerone: sorry, I got a merge conflict and must have chosen the wrong order.

Schparky: Oh, that makes sense. No problem.

application/domain/domain.go
Schparky: Gotcha. Sounds right.

application/listeners/posthelpers.go
Baggerone: Exported - done.

application/models/post.go
Baggerone: I think I got it corrected in the latest push

application/listeners/listeners.go
Baggerone: Sounds good.

application/domain/domain.go
```diff
@@ -41,15 +41,32 @@ const (
...
+	EventApiMessageCreated    = "api:message:created"
+	EventApiPostStatusUpdated = "api:post:status:updated"
 )
 
 // Notification Message Template Names
 const (
-	MessageTemplateNewMessage = "new_message"
-	MessageTemplateNewRequest = "new_request"
+	MessageTemplateNewMessage                      = "new_message"
+	MessageTemplateNewRequest                      = "new_request"
```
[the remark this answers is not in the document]
Baggerone: That was already there from Schram's changes, but I believe it's referring to a Post of Type Request (vs. Offer).

PR #86, opened 2019-10-22
Use workers to delay "New Message" notification

Schparky:


application/job/job.go
```diff
@@ -0,0 +1,100 @@
...
+		if p.ID == m.SentBy.ID {
+			continue
+		}
+
+		var tp models.ThreadParticipant
+		if err := tp.FindByThreadIDAndUserID(m.ThreadID, p.ID); err != nil {
+			return err
+		}
+		// Don't send a notification if this user has viewed the message or if they've already been notified
+		if tp.LastViewedAt.After(m.UpdatedAt) || tp.LastNotifiedAt.After(m.UpdatedAt) {
```
fillup: Does`tp.LastNotifiedAt.After(m.UpdatedAt)` mean if a notification was sent to me yesterday, and a new message is posted in the thread today that I will not receive a new notification today because I was notified yesterday?

Schparky: No, this is looking at the `updated_at` for the message that triggered the event. So this is only preventing a second notification for a given message. In other words, if two messages have been posted since the last notification, only the first will trigger a notification.

application/job/job.go
```diff
@@ -0,0 +1,100 @@
...
+		}
+
+		msg.ToName = p.Nickname
+		msg.ToEmail = p.Email
+		if err := notifications.Send(msg); err != nil {
+			domain.ErrLogger.Printf("error sending 'New Message' notification, %s", err)
+			continue
+		}
+
+		if err := tp.UpdateLastNotifiedAt(time.Now()); err != nil {
```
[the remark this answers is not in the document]
Schparky: The `continue` a couple lines above here should take care of that.

PR #88, opened 2019-10-23
add thread lastViewedAt to GraphQL schema

Schparky:


wcjr:
I'm wondering if we'll want some sort of roll-up of "unread" messages at the user level at some point...not sure yet.

PR #94, opened 2019-10-25
unread message count on User and Thread

Baggerone:


application/gqlgen/thread.go
```diff
@@ -72,6 +72,23 @@ func (r *threadResolver) Post(ctx context.Context, obj *models.Thread) (*models.
 	return obj.GetPost(selectedFields)
 }
 
+func (r *threadResolver) UnreadMessageCount(ctx context.Context, obj *models.Thread) (int, error) {
+	if obj == nil {
+		return 0, nil
+	}
+	lastViewedAt, err := obj.GetLastViewedAt(models.GetCurrentUserFromGqlContext(ctx, TestUser))
+	if err != nil {
```
Schparky:
```suggestion
	if err != nil || lastViewedAt == nil {
```

application/models/thread.go
```diff
@@ -197,3 +197,20 @@ func (t *Thread) Load(fields ...string) error {
...
+func (t *Thread) UnreadMessageCount(lastViewedAt time.Time) (int, error) {
+	count := 0
+
+	msgs, err := t.GetMessages([]string{"created_at"})
+	if err != nil {
+		return count, err
+	}
+
+	for _, m := range msgs {
+		if m.CreatedAt.After(lastViewedAt) {
```
Schparky: Wondering if this should be `UpdatedAt`. Doesn't matter today, but I'm curious about what happens if we allow editing of old messages.

application/models/thread_fixtures_test.go
```diff
@@ -0,0 +1,252 @@
...
+		Name:       fmt.Sprintf("ACME-%s", unique),
+		Uuid:       domain.GetUuid(),
+		AuthType:   AuthTypeSaml,
+		AuthConfig: "{}",
+	}
+
+	if err := ms.DB.Create(&org); err != nil {
+		t.Errorf("error creating org %+v ...\n %v \n", org, err)
+		t.FailNow()
+	}
```
Schparky: `createFixture(&org)`

application/models/user_fixtures_test.go
```diff
@@ -0,0 +1,276 @@
...
+			UserID:       posts[1].ProviderID.Int,
+			LastViewedAt: futureTime,
+		},
+	}
+
+	for i := range threadParticipants {
+		createFixture(ms, &threadParticipants[i])
+	}
+
+	// I can't seem to give them custom times
```
Schparky: I had to do raw SQL inserts in `job_fixtures_test.go`. e.g.:
```go
		err := js.DB.RawQuery(`INSERT INTO messages (content, created_at, sent_by_id, thread_id, updated_at, uuid)
			VALUES (?, ?, ?, ?, ?, ?)`, m.Content, m.UpdatedAt, m.SentByID, m.ThreadID, m.UpdatedAt, m.Uuid).Exec()
```
It seems like you solved this by using future `last_viewed_at` times. That's fine with me.

application/models/thread_fixtures_test.go
Baggerone: `createFixture(ms, &org)`   :-)

application/models/thread.go
Baggerone: I checked with Billy.  He doesn't think we will end up allowing users to edit messages.

application/gqlgen/thread.go
Baggerone: Good catch.  I've added a separate `if` block for it now to log a different message.

application/models/user_fixtures_test.go
Baggerone: It's a shame they don't let you force the CreatedAt value.  (Maybe they do give a way to make that happen?)

PR #96, opened 2019-10-25
remove myThreadID

Schparky:
According to @forevermatt, this field is no longer used.


Schparky:
Previous branch hasn't been merged yet, so this looks messy.

PR #97, opened 2019-10-25
on creating a new message update LastViewedAt to now

Baggerone:


application/models/message.go
```diff
@@ -75,6 +75,30 @@ func (m *Message) ValidateUpdate(tx *pop.Connection) (*validate.Errors, error) {
...
 
+// AfterCreate updates the LastViewedAt value on the associated ThreadParticipant to right now
+func (m *Message) AfterCreate(tx *pop.Connection) error {
+	threadP := ThreadParticipant{}
+	where := "thread_id = ? AND user_id = ?"
+	if err := DB.Where(where, m.ThreadID, m.SentByID).First(&threadP); err != nil {
+		domain.ErrLogger.Printf("error finding thread participant for Thread ID %v and User ID %v",
+			m.ThreadID, m.SentByID)
+		return nil
+	}
```
Schparky: Can use `FindByThreadIDAndUserID` here.

application/models/message_fixtures_test.go
```diff
@@ -180,3 +184,204 @@ func Fixtures_Message_FindByUUID(ms *ModelSuite) MessageFixtures {
...
+
+	// Because of Message.AfterCreate() we need to reset the LastViewedAt attributes
+	//                     Eager User0, Lazy User1, Lazy User1, Eager User0
+	for i, tt := range []time.Time{tNow, oldTime, oldTime, tNow} {
+		if err := threadParticipants[i].UpdateLastViewedAt(tt); err != nil {
+			t.Errorf("could not update LastViewedAt ... %v", err)
+			t.FailNow()
+		}
+	}
+	println()
```
Schparky:
```suggestion

```

application/models/message_fixtures_test.go
```diff
@@ -180,3 +184,204 @@ func Fixtures_Message_FindByUUID(ms *ModelSuite) MessageFixtures {
...
+	for _, m := range messages {
+		if err := ms.DB.RawQuery(`INSERT INTO messages (content, created_at, sent_by_id, thread_id, updated_at, uuid)`+
+			`VALUES (?, ?, ?, ?, ?, ?)`,
+			m.Content, m.CreatedAt, m.SentByID, m.ThreadID, m.CreatedAt, m.Uuid).Exec(); err != nil {
+			t.Errorf("error loading messages ... %v", err)
+			t.FailNow()
+		}
+	}
+
+	// Because of Message.AfterCreate() we need to reset the LastViewedAt attributes
```
Schparky: Really? I'm surprised that `RawQuery` triggers `AfterCreate`.

application/models/thread.go
```diff
@@ -202,7 +202,9 @@ func (t *Thread) Load(fields ...string) error {
 	return nil
 }
 
-func (t *Thread) UnreadMessageCount(lastViewedAt time.Time) (int, error) {
+// UnreadMessageCount returns the number of messages on this thread that the current
+//  user has not created and for which the CreatedAt value is after the lastViewedAt value
+func (t *Thread) UnreadMessageCount(user *User, lastViewedAt time.Time) (int, error) {
```
Schparky: I don't like that `user` is a pointer. It indicates that the function could modify it. If it's for performance reasons, I suspect that optimizations eliminate any differences.

application/models/message.go
Baggerone: Good catch.

application/models/message_fixtures_test.go
Baggerone: Done

application/models/thread.go
Baggerone: Sounds good.

application/models/message_fixtures_test.go
Baggerone: Oops, I should have taken out that code after I switched from using pop to RawQuery.

PR #98, opened 2019-10-28
address TODOs

Schparky:
- added stub for Offer status validation
- left a placeholder for Org logo, but removed TODO

### DIFF
--- a/application/actions/auth.go
+++ b/application/actions/auth.go
@@ -124,7 +124,7 @@ func provideOrgOptions(userOrgs models.UserOrganizations, c buffalo.Context) err
 		orgOpts = append(orgOpts, AuthOrgOption{
 			ID:      strconv.Itoa(uo.ID),
 			Name:    uo.Organization.Name,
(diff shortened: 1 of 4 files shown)


application/actions/auth.go
```diff
@@ -124,7 +124,7 @@ func provideOrgOptions(userOrgs models.UserOrganizations, c buffalo.Context) err
 		orgOpts = append(orgOpts, AuthOrgOption{
 			ID:      strconv.Itoa(uo.ID),
 			Name:    uo.Organization.Name,
-			LogoURL: uo.Organization.Url.String, // TODO change to a logo url when one is added to organization
+			LogoURL: "",
```
Baggerone: Shouldn't this still have a TODO similar to the original one?

Schparky: Yes, if we want to use TODO comments. At this point, everything should be managed on the Trello board. I'll make sure Org Logos are listed there.

PR #102, opened 2019-10-29
More notifications and associated tidying

Baggerone:


application/listeners/posthelpers.go
```diff
@@ -63,7 +63,73 @@ func sendNotificationRequestFromOpenToCommitted(template string, post m.Post) {
 	}
 }
 
-func sendNotificationRequestFromCommittedToAccepted(template string, post m.Post) {
+// Until we have status auditing history, we don't know who reverted the Post to `open` status.
```
Schparky:
```suggestion
// Until we have status auditing history, we don't know who reverted the Post to `accepted` status.
```

application/listeners/posthelpers.go
```diff
@@ -63,7 +63,73 @@ func sendNotificationRequestFromOpenToCommitted(template string, post m.Post) {
...
+		"requesterNickname": postUsers.Requester.Nickname,
+		"requesterEmail":    postUsers.Requester.Email,
+	}
+
+	msg = notifications.Message{
+		Template: template,
+		Data:     data,
+		ToName:   oldProvider.Nickname,
+		ToEmail:  oldProvider.Email,
+	}
```
Schparky: Does this all need to be repeated? Maybe just change the fields that need to be changed.

application/listeners/posthelpers.go
Baggerone: Heh, heh, I think the UI threw you off. You're suggestion appears to be based on the previous function name rather than the new one.

application/listeners/posthelpers.go
Baggerone: Good catch

PR #109, opened 2019-11-12
Add "isEditable" flag to Post

Schparky:
- adds to GraphQL schema so UI can easily know whether to show an "edit" button
- blocks `updatePost` when not appropriate

### DIFF
--- a/application/gqlgen/post.go
+++ b/application/gqlgen/post.go
@@ -363,6 +363,16 @@ func (r *mutationResolver) UpdatePost(ctx context.Context, input postInput) (*mo
 		return nil, err
 	}
 
+	var editable bool
+	if editable, err = post.IsEditable(cUser); err != nil {
+		domain.Error(models.GetBuffaloContextFromGqlContext(ctx), err.Error())
+		return nil, err
+	}
+	if !editable {
+		domain.Error(models.GetBuffaloContextFromGqlContext(ctx), "post cannot be edited")
+		return nil, err
+	}
+
 	if err := post.Update(); err != nil {
 		domain.Error(models.GetBuffaloContextFromGqlContext(ctx), err.Error())
 		return nil, err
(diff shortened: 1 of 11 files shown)


application/gqlgen/post.go
```diff
@@ -260,6 +261,15 @@ func (r *postResolver) Files(ctx context.Context, obj *models.Post) ([]models.Fi
 	return files, nil
 }
 
+// IsEditable response with true if the current user is the creator of the post, and it is not in a locked status.
```
wcjr: comment doesn't seem necessary here and it will need to be maintained should the logic down in `IsEditable` ever changes, I'd recommend losing this one.

application/models/post.go
```diff
@@ -549,3 +549,46 @@ func (p *Post) SetOrigin(location Location) error {
...
+	}
+
+	return p.isStatusEditable(p.Status), nil
+}
+
+// isStatusEditable defines which posts statuses can be edited. It does not use the receiver `p'.
+func (p *Post) isStatusEditable(status string) bool {
+	switch status {
+	case PostStatusOpen:
+		fallthrough
```
wcjr: wow, really?!  This is gross, sorry you have to do switches like this in Go :-D

application/models/user.go
```diff
@@ -246,6 +246,11 @@ func (u *User) CanEditOrganization(orgId int) bool {
 	return false
 }
 
+// CanEditAllPosts returns true if the given user is allowed to edit any post
```
wcjr: again, this comment is creating more confusion than the code itself (imo)

application/gqlgen/post.go
Schparky: Good point.

application/models/post.go
Schparky: Are you referring to the `fallthrough` keyword? I seem to use `break` more in traditional languages that I use `fallthrough` in Go. I like the change.

application/models/user.go
Schparky: Perhaps. The linter gives a warning for undocumented, exported functions, and I like to clear up those warnings. I think something is better than nothing, in terms of function docs.

PR #111, opened 2019-11-13
make destination not nullable in database

Schparky:
The change to the API to make destination not nullable was made earlier, but that resulted in the potential for an error condition if a post had no destination.


application/models/location.go
```diff
@@ -22,3 +27,68 @@ func (l Location) String() string {
...
+	Latitude, Longitude nulls.Float64
+}
+
+// IsValid checks the latitude and longitude valid ranges
+func (v *geoValidator) IsValid(errors *validate.Errors) {
+	if !v.Latitude.Valid && !v.Longitude.Valid {
+		return
+	}
+
+	if v.Latitude.Valid != v.Latitude.Valid {
```
Baggerone:
```suggestion
	if v.Latitude.Valid != v.Longitude.Valid {
```

Schparky: Good catch!

PR #116, opened 2019-11-19
Translated email subjects 

Baggerone:
Sorry, I had to put the post status transitions in alphabetical order.  I was struggling with getting things set up in the yaml localization file with the order I had them in originally.


application/listeners/posthelpers.go
```diff
@@ -2,12 +2,29 @@ package listeners
...
 	"github.com/silinternational/wecarry-api/models"
 	"github.com/silinternational/wecarry-api/notifications"
 )
 
+// TODO create a proper solution for this
+func getUserLanguage() string {
+	return "en"
+}
+
+type argAppName struct{ AppName string }
```
Schparky: Will this work as a map[string]string as well?

application/listeners/posthelpers.go
```diff
@@ -2,12 +2,29 @@ package listeners
...
+
+func getTranslatedSubject(translationID, template string) string {
+	subj, err := domain.TranslateWithLang(getUserLanguage(), translationID, argAppName{AppName: domain.Env.AppName})
```
Schparky: Would it make sense to include the `AppName` arg in the `TranslateWithLang` function? Might be an arg that is needed more often than not.

application/notifications/messages.go
```diff
@@ -9,4 +9,5 @@ type Message struct {
 	ToName    string
 	ToEmail   string
 	ToPhone   string
+	Subject   string
```
Schparky: This field isn't read anywhere. Looks like `(e *EmailNotifier) Send` function needs to pass it on to the email service.

application/domain/domain.go
```diff
@@ -347,3 +352,21 @@ func GetPostUIURL(postUUID string) string {
 func GetThreadUIURL(threadUUID string) string {
 	return Env.UIURL + threadUIPath + threadUUID
 }
+
+// TODO watch for when this is available on a new version of mw-i18n and then replace calls to this function
+// TranslateWithLang returns the translation of the string identified by translationID, for the given language.
+// See Translate for further details.
```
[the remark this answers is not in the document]
Baggerone: Hopefully, we won't need a separate function like this once they have the original released.

application/listeners/posthelpers.go
Baggerone: Hopefully the `TranslateWithLang` function will be getting replaced by a near-future upgrade.

application/notifications/messages.go
Baggerone: Good catch.

PR #117, opened 2019-11-20
User preferences

Baggerone:
Small PR to make sure I'm on the right track.


application/domain/domain.go
```diff
@@ -75,6 +75,20 @@ const (
...
+const (
+	UserPreferenceKeyLanguage        = "language"
+	UserPreferenceLanguageEnglish    = "en"
+	UserPreferenceLanguageFrench     = "fr"
+	UserPreferenceLanguageSpanish    = "es"
+	UserPreferenceLanguageKorean     = "ko"
+	UserPreferenceLanguagePortuguese = "pt"
+
+	UserPreferenceKeyUnits      = "units"
+	UserPreferenceUnitsImperial = "imperial"
```
Schparky:
```suggestion
	UserPreferenceUnitsUSCustomary = "uscs"
```
Since UK and Canada primarily use metric, US customary is probably more applicable.

application/domain/domain.go
```diff
@@ -370,3 +384,17 @@ func TranslateWithLang(lang, translationID string, args ...interface{}) (string,
 	}
 	return t2(translationID, args...), nil
 }
+
+// IsErrorMoreThanJustNoSQLRows returns false if the error is nil or is just reporting that there
+//   were no rows in the result set for a sql query.
+func IsErrorMoreThanJustNoSQLRows(err error) bool {
```
Schparky: Somewhat shorter suggestion: `IsErrorOtherThanNoRows`

application/models/user.go
```diff
@@ -495,3 +496,27 @@ func (u *User) WantsPostNotification(post Post) bool {
...
+}
+
+// GetPreference returns a pointer to a matching UserPreference record or if
+// none is found, returns nil
+func (u *User) GetPreference(key string) (*UserPreference, error) {
+	uPref := UserPreference{}
+
+	err := DB.Where("user_id = ?", u.ID).Where("key = ?", key).First(&uPref)
+	if err != nil {
+		if domain.IsErrorMoreThanJustNoSQLRows(err) {
```
Schparky: Since the function returns `false` for `err == nil`, do you need both `if`s?

application/domain/domain.go
Schparky: Or just `IsOtherThanNoRows`

Baggerone: Sounds good.

application/models/user.go
Baggerone: I think so, since if the error is not nil and not `NoRows` we return nil, err,
  if the error is not nil but is 'NoRows we return nil, nil, 
  and if the error is not nil, we return the object and nil.

Do you see a way of getting those three different things done with only one `if`?

application/domain/domain.go
Baggerone: As long as the UI uses well known terminology.  I don't remember hearing of USCustomary before.

application/models/user.go
Schparky: Would this work?
```go
if domain.IsErrorMoreThanJustNoSQLRows(err) {
    return nil, err
}
return &uPref, nil
```
Since the return value is probably nil when there's a no-rows error, or any error for that matter, this might be the same as what you have.

Schparky: Oh, I think I see what I missed before. You want to return a nil `*UserPreference` rather than a pointer to an empty `UserPreference` when there were no rows.

PR #118, opened 2019-11-20
location-based notification for new posts

Schparky:


application/models/location.go
```diff
@@ -92,3 +95,36 @@ func (l *Location) Create() error {
...
+
+	var a = 0.5 - math.Cos((lat2-lat1)*p)/2 +
+		math.Cos(lat1*p)*math.Cos(lat2*p)*
+			(1-math.Cos((lon2-lon1)*p))/2
+
+	return 12742 * math.Asin(math.Sqrt(a)) // 2 * R; R = 6371 km
+}
+
+// IsNear answers the question "Are these two locations near each other?"
+func (l *Location) IsNear(l2 Location) bool {
```
Baggerone: How about `loc2`, so it doesn't look like `twelve` on line 125

application/models/location.go
```diff
@@ -92,3 +95,36 @@ func (l *Location) Create() error {
 
 	return nil
 }
+
+// DistanceKm calculates the distance in km between two locations
```
Baggerone: Maybe note that it's from that SO answer

PR #123, opened 2019-11-22
some gqlgen tests and a models.Message fix

Baggerone:


application/gqlgen/message_test.go
```diff
@@ -0,0 +1,80 @@
...
+	var resp messageResponse
+
+	TestUser = f.Users[0]
+	err := c.Post(query, &resp)
+	gs.NoError(err)
+
+	thread, err := f.Messages[0].GetThread([]string{"uuid", "id"})
+	gs.NoError(err)
+
+	participants, err := thread.GetParticipants([]string{"nickname"})
```
Schparky: What is the reason to test second- and third-level properties? Seems they will be tested in other places.

application/gqlgen/thread_test.go
```diff
@@ -0,0 +1,113 @@
...
+
+	TestUser = f.Users[0]
+	err := c.Post(query, &resp)
+
+	gs.NoError(err)
+
+	gs.Equal(f.Threads[0].Uuid.String(), resp.Threads[0].ID)
+	gs.Equal(f.Posts[0].Uuid.String(), resp.Threads[0].Post.ID)
+	gs.Equal(f.Messages[0].Uuid.String(), resp.Threads[0].Messages[0].ID)
+	gs.Equal(f.Messages[0].Content, resp.Threads[0].Messages[0].Content)
```
Schparky: Same thing here about data depth.

application/models/message.go
```diff
@@ -76,7 +76,24 @@ func (m *Message) ValidateUpdate(tx *pop.Connection) (*validate.Errors, error) {
 }
 
 // AfterCreate updates the LastViewedAt value on the associated ThreadParticipant to right now
+// It also ensures the associated ThreadParticipant records exist
 func (m *Message) AfterCreate(tx *pop.Connection) error {
+
+	thread, err := m.GetThread([]string{"uuid", "id"})
+	if err != nil {
+		return errors.New("error getting message's Thread ... " + err.Error())
```
Schparky: Side note: this may be a case where the insert should be rolled back.

application/models/thread.go
```diff
@@ -147,6 +139,39 @@ func (t *Thread) CreateWithParticipants(postUuid string, user User) error {
 	}
 
 	*t = thread
+	return t.ensureParticipants(post, user.ID)
+}
+
+func (t *Thread) ensureParticipants(post Post, userID int) error {
```
Schparky: Would it be more possible to do something like this?
```go
var tp ThreadParticipant
if err := t.FindByThreadIDAndUserID(t.ID, userID); domain.IsOtherThanNoRows(err) {
    tp.ThreadID = t.ID
    tp.UserID = userID
    tp.Save()
}
```

Schparky: `ensureParticipants` needs a test

application/models/thread.go
```diff
@@ -147,6 +139,39 @@ func (t *Thread) CreateWithParticipants(postUuid string, user User) error {
...
+	for _, uID := range []int{post.CreatedByID, userID} {
+		if err := t.createParticipantIfNeeded(threadParticipants, uID); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (t *Thread) createParticipantIfNeeded(tpUsers Users, userID int) error {
```
Schparky: `createParticipantIfNeeded` needs a test

application/gqlgen/message_test.go
Baggerone: Basically, I don't trust that just because they give good results as top level requests, they will give good results as nested requests (based on experience).

application/models/thread.go
Baggerone: That doesn't appear to get the Post creator included as a participant.

Schparky: OK. I guess that chunk would have to be repeated for the creator.

application/models/thread.go
Baggerone: I pulled this code out of ensure Participants early on to make it more readable (using return statements, instead of extra `if`s.  It's not exported and gets tested by the tests for ensure Participants and CreateWithParticipants.

I think that should be good enough, don't you?

application/models/thread.go
Baggerone: Done

Baggerone: Also, we couldn't do it exactly the way you have it, since `t.FindByThreadIDAndUserID` doesn't return the same kind of error that `IsOtherThanNoRows` is expecting.

application/models/thread.go
Schparky: In general, I'd advocate for testing the lower-level function (end of the call chain) more than the higher-level, because it requires fewer permutations to ensure coverage of all possible scenarios. In this case, there is only one caller to `createParticipantIfNeeded`, so I think your test coverage is 100%. Yes, it's good as it is.

application/models/thread.go
Schparky: Good point, `IsOtherThanNoRows` would have to be changed to look for "sql: no rows in result set" within the error message, rather than an exact compare.

PR #125, opened 2019-11-25
Remove select fields

Baggerone:


application/gqlgen/user.go
```diff
@@ -171,7 +167,7 @@ func (r *queryResolver) User(ctx context.Context, id *string) (*models.User, err
 
 	dbUser := models.User{}
 	selectFields := GetSelectFieldsForUsers(ctx)
```
Schparky: `selectFields` can be removed here if you don't need the `extras` below. Then `GetSelectFieldsForUsers` and `UserFields` can go away as well.

PR #126, opened 2019-11-25
Remove gravatar URL creation

Schparky:
We decided it is better to handle gravatar at the UI so it can decide on resolution, etc. Renamed database column to avoid confusion with GQL photoURL field.

### DIFF
--- a/application/migrations/schema.sql
+++ b/application/migrations/schema.sql
@@ -513,7 +513,7 @@ CREATE TABLE public.users (
     updated_at timestamp without time zone NOT NULL,
     uuid uuid NOT NULL,
     photo_file_id integer,
-    photo_url character varying(255),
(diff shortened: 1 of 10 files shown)


application/migrations/schema.sql
```diff
@@ -513,7 +513,7 @@ CREATE TABLE public.users (
     updated_at timestamp without time zone NOT NULL,
     uuid uuid NOT NULL,
     photo_file_id integer,
-    photo_url character varying(255),
+    auth_photo_url character varying(255),
```
wcjr: is this short for "Author" or "Authentication source"?

Schparky: Is that a rhetorical question? It's short for 'authentication source'.

PR #127, opened 2019-11-26
enforce message privacy

Schparky:
Only thread participants and super admins can see messages


application/models/message.go
```diff
@@ -192,3 +192,24 @@ func (m *Message) FindByUUID(id string) error {
...
+	if user.AdminRole == UserAdminRoleSuperAdmin {
+		return nil
+	}
+
+	var tp ThreadParticipant
+	if err := tp.FindByThreadIDAndUserID(m.ThreadID, user.ID); err != nil {
+		if domain.IsOtherThanNoRows(err) {
+			return fmt.Errorf("error finding threadParticipant record for message %s, %s", id, err)
+		}
+		return fmt.Errorf("user %s has insufficient permissions to read message %s", user.Uuid.String(), id)
```
Baggerone: I'm thinking that if this message doesn't get back out to the UI, it might be better to use the user.ID field, since the UUID is so long.  (I guess that could apply to the message id as well throughout this function)

If you want to keep the UUID field, you can change the first `%s` to `%v` and get rid of `.String()`.

application/models/message.go
```diff
@@ -192,3 +192,24 @@ func (m *Message) FindByUUID(id string) error {
 
 	return nil
 }
+
+// FindByUserAndUUID loads from DB the Message record identified by the given UUID, if the given user is allowed.
+func (m *Message) FindByUserAndUUID(user User, id string) error {
+	if err := m.FindByUUID(id); err != nil {
```
Baggerone: I suppose`FindByUUID` shouldn't be exported anymore, so we don't accidentally use it instead of `FindByUserAndUUID`.

application/models/message.go
Schparky: I think I'll continue with the UUID since we have some ability to troubleshoot via raw graphql and possibly, eventually, through an admin interface. I just played with it and `%s` works without the `.String()` as well. I was thinking `%v` would print it as a byte array, but I may have been thinking of `%#v`.

PR #130, opened 2019-12-03
Add post status history table and allow back tracking

Baggerone:


application/models/posthistory_fixtures_test.go
```diff
@@ -0,0 +1,63 @@
+package models
+
+import (
+	"github.com/silinternational/wecarry-api/domain"
+	"strconv"
+
+	"github.com/gobuffalo/nulls"
```
Schparky:
```suggestion
	"github.com/gobuffalo/nulls"
	"github.com/silinternational/wecarry-api/domain"
```

application/models/posthistory_fixtures_test.go
```diff
@@ -0,0 +1,63 @@
+package models
+
+import (
+	"github.com/silinternational/wecarry-api/domain"
```
Schparky:
```suggestion

```

application/models/post.go
```diff
@@ -745,3 +823,59 @@ func (p *Post) GetLocationForNotifications() (*Location, error) {
 	}
 	return &postLocation, nil
 }
+
+// createNewHistory checks if the post has a status that is different than the
+// most recent of its Post History entries.  If so, it creates a new Post History
+// with the Post's new status.
+func (p *Post) createNewHistory() error {
```
Schparky: Should this be a `PostHistory` function: `func (p *PostHistory) New(post Post) error`?

application/models/post.go
```diff
@@ -745,3 +823,59 @@ func (p *Post) GetLocationForNotifications() (*Location, error) {
...
+			return err
+		}
+	}
+
+	return nil
+}
+
+// popHistory deletes the most recent postHistory entry for a post
+// assuming it's status matches the expected one.
+func (p *Post) popHistory(currentStatus PostStatus) error {
```
Schparky: Same here? `func (p *PostHistory) Pop(postID int, currentStatus PostStatus) error`

application/models/post_fixtures_test.go
```diff
@@ -43,7 +44,7 @@ func CreateFixturesValidateUpdate(ms *ModelSuite, t *testing.T) []Post {
 		t.FailNow()
 	}
 
-	locations := []Location{{}, {}, {}, {}, {}, {}}
+	locations := []Location{{}, {}, {}, {}, {}, {}, {}}
```
Schparky:
```suggestion
	locations := make([]Location, 7)
```

application/models/post_test.go
```diff
@@ -416,6 +418,14 @@ func (ms *ModelSuite) TestPost_ValidateUpdate() {
 			},
 			wantErr: false,
 		},
+		{
```
Schparky: If you're so inclined, I think this test (`TestPost_ValidateUpdate`) might be better (e.g. better coverage, smaller fixtures, easier to read) as specific tests on each validator. Sorry I didn't think of it earlier, and no sweat if it's not worth the trouble now.

application/models/posthistory_fixtures_test.go
Baggerone: Done

application/models/post.go
Baggerone: That seems less intuitive to me, since you don't start with a PostHistory that you're doing something with. Rather, you have a Post for which you want to create a new related PostHistory.

application/models/post_fixtures_test.go
Baggerone: Good change

application/models/post.go
Schparky: I guess you could look at it that way. My thought was that almost every line in the function is operating on a PostHistory. Also, Post is becoming a "god class" and I think we need to be on the lookout for things that don't belong there.

application/models/post_test.go
Baggerone: Done

application/models/post.go
Baggerone: I'll want to get this merged for Billy's sake.  I'll make that change in the next one.

PR #132, opened 2019-12-05
Restrict user profile data visibility

Schparky:
Change GraphQL schema to only return a full user profile from the `user` and `users` queries and `updateUser` mutation, which have restricted access. This is a breaking change to the API.

This requires that the Gravatar URL generation be moved back into the API.

### DIFF
--- a/application/gqlgen/user.go
+++ b/application/gqlgen/user.go
@@ -62,8 +62,8 @@ func (r *userResolver) Posts(ctx context.Context, obj *models.User, role PostRol
 	return posts, nil
 }
 
-// PhotoURL retrieves a URL for the user profile photo or avatar.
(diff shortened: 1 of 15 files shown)


application/gqlgen/user.go
```diff
@@ -234,3 +234,16 @@ func (r *userResolver) Preferences(ctx context.Context, obj *models.User) (*mode
 
 	return &standardPrefs, nil
 }
+
+// getPublicProfile converts a models.User to a PublicProfile, which hides private profile information
+func getPublicProfile(user *models.User) (*PublicProfile, error) {
+	url, err := user.GetPhotoURL()
+	if err != nil {
+		return nil, err
```
Baggerone: Just wondering if it would be better to only log the error and  not return an error here.

PR #133, opened 2019-12-05
Also don't provide completed posts

Baggerone:
Billy requested that we don't include `COMPLETED` posts out to the UI in our lists of Posts.

### DIFF
--- a/application/models/post.go
+++ b/application/models/post.go
@@ -659,7 +659,7 @@ func scopeNotRemoved() pop.ScopeFunc {
 	}
 }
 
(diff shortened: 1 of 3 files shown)


application/models/post.go
```diff
@@ -659,18 +659,26 @@ func scopeNotRemoved() pop.ScopeFunc {
 	}
 }
 
+// scope query to not include removed posts
```
Schparky:
```suggestion
// scope query to not include removed or completed posts
```

PR #136, opened 2019-12-06
Feature/adds kilograms to posts

Baggerone:
I made kilograms non-nullable in the db, since I assume we can interpret "no weight recorded" and "zero weight" as the same thing.

Also, I chose "kilograms" because I figured it is less ambiguous than "pounds".


application/gqlgen/schema.graphql
```diff
@@ -137,6 +137,7 @@ type Post {
     updatedAt: Time!
     url: String
     cost: String
+    kilograms: String
```
wcjr: `String`?  Wouldn't a float or number be more appropriate?

fillup: When we get to conversions I'd expect to have some decimal numbers, so float might be better for that from the gql schema portion

Baggerone: I was mimicking how we did it for `cost`, but I'll make it a float now.

PR #140, opened 2019-12-09
use users language for translation

Baggerone:
I also replaced a couple of series of `if` blocks with maps and structs.
And, I replaced "requester" with "receiver"


application/listeners/posthelpers.go
```diff
@@ -9,15 +9,10 @@ import (
...
-
 var argAppName = map[string]string{"AppName": domain.Env.AppName}
 
-func getTranslatedSubject(translationID, template string) string {
-	subj, err := domain.TranslateWithLang(getUserLanguage(), translationID, argAppName)
+func getTranslatedSubject(language, translationID, template string) string {
+	subj, err := domain.TranslateWithLang(language, translationID, argAppName)
 
 	if err != nil {
 		domain.ErrLogger.Printf("error translating '%s' notification subject, %s", template, err)
```
Schparky: Just a thought: what if you logged the subject itself, and not pass in the template?

application/listeners/posthelpers_fixtures_test.go
```diff
@@ -62,6 +62,13 @@ func CreateFixtures_GetPostRecipients(ms *ModelSuite, t *testing.T) orgUserPostF
 		}
 	}
 
+	_, err := users[1].UpdateStandardPreferences(models.StandardPreferences{Language: domain.UserPreferenceLanguageFrench})
+
+	if err != nil {
+		t.Errorf("could not create language preference for user %v ... %v.", users[1].Nickname, err)
```
Schparky:
```suggestion
		ms.NoErr(err, "could not create language preference for user %v ... %v.", users[1].Nickname, err)
```
and remove the `if` and `FailNow()` here and elsewhere in this function.

application/models/user.go
```diff
@@ -550,132 +550,55 @@ func (u *User) GetPreferences() (StandardPreferences, error) {
...
+	for fieldName, fV := range fieldAndValidators {
+		if value, ok := dbPreferences[fieldName]; ok {
+			if fV.validator(value) {
+				finalValues[fieldName] = value
+			} else {
+				domain.Logger.Printf("user preference %s in database not allowed ... %s", fieldName, value)
+			}
 		}
-		return nil, nil
 	}
```
Schparky: Can this section be moved into a separate function in `userpreference.go`?

application/listeners/posthelpers.go
Baggerone: Do you mean, log the `translationID`?

Schparky: Yes

application/listeners/posthelpers_fixtures_test.go
Baggerone: Good catch

PR #141, opened 2019-12-09
better email layout

Schparky:


application/domain/domain.go
```diff
@@ -451,3 +451,21 @@ func GetStructTags(tagType string, s interface{}) (map[string]string, error) {
...
+
+	if err != nil {
+		ErrLogger.Printf("error translating '%s' notification subject, %s", template, err)
+	}
+
+	return subj
+}
+
+// This is intended as a temporary fill-in
+func getUserLanguage() string {
```
Baggerone: I think this isn't used anymore.

application/domain/domain.go
```diff
@@ -451,3 +451,21 @@ func GetStructTags(tagType string, s interface{}) (map[string]string, error) {
 
 	return fieldTags, nil
 }
+
+func GetTranslatedSubject(translationID, template string) string {
```
Baggerone: This function got changed as well as its signature.

PR #144, opened 2019-12-10
move Post.BeforeUpdate code to Post.AfterUpdate

Baggerone:


application/models/posthistory.go
```diff
@@ -114,3 +114,14 @@ func (pH PostHistory) popForPost(post Post, currentStatus PostStatus) error {
 
 	return nil
 }
+
+// pop deletes the most recent postHistory entry for a post
+// assuming it's status matches the expected one.
```
Schparky: Comment is incorrect.

PR #148, opened 2019-12-12
Add a gql endpoint to search for posts with a certain string in the Title or Description

Baggerone:


application/actions/post_fixtures_test.go
```diff
@@ -160,6 +160,58 @@ func createFixturesForPostQuery(as *ActionSuite) PostQueryFixtures {
...
+		userOrgs[i].AuthID = unique + "_auth_user" + strconv.Itoa(i)
+		userOrgs[i].AuthEmail = unique + users[i].Email
+		createFixture(as, &userOrgs[i])
+
+		accessTokenFixtures[i].UserID = users[i].ID
+		accessTokenFixtures[i].UserOrganizationID = userOrgs[i].ID
+		accessTokenFixtures[i].AccessToken = models.HashClientIdAccessToken(users[i].Nickname)
+		accessTokenFixtures[i].ExpiresAt = time.Now().Add(time.Minute * 60)
+		createFixture(as, &accessTokenFixtures[i])
+	}
```
Schparky: This can be replaced with `test.CreateUserFixtures(as.DB, as.T(), 2)` eventually.

application/actions/post_test.go
```diff
@@ -289,3 +293,21 @@ func (as *ActionSuite) Test_UpdatePostStatus() {
 		}
 	}
 }
+
+func (as *ActionSuite) Test_SearchRequests() {
+	f := createFixturesForSearchRequestsQuery(as)
+	query := `{ searchRequests(text: "match")
```
Schparky: If you change this, you can use `PostsResponse` rather than `SearchRequestsResponse`
```suggestion
	query := `{ posts: searchRequests(text: "match")
```

Baggerone: Good catch.

PR #154, opened 2019-12-17
add `eventID` to `createPost` mutation

Schparky:


application/gqlgen/schema.graphql
```diff
@@ -211,6 +211,7 @@ input CreatePostInput {
     url: String
     kilograms: Float
     photoID: ID
+    eventID: ID
```
Baggerone: This should be meetingID, right?

application/gqlgen/post.go
```diff
@@ -311,18 +321,18 @@ func (r *mutationResolver) CreatePost(ctx context.Context, input postInput) (*mo
 	}
 
 	dest := convertGqlLocationInputToDBLocation(*input.Destination)
-	if err1 := dest.Create(); err1 != nil {
-		return nil, reportError(ctx, err1, "CreatePost.SetDestination", extras)
+	if err = dest.Create(); err != nil {
```
Baggerone: Thanks for changing  these.

application/actions/post_test.go
```diff
@@ -235,6 +236,7 @@ func (as *ActionSuite) Test_CreatePost() {
 
 	as.Equal(f.Organization.UUID.String(), postsResp.Post.Organization.ID)
 	as.Equal(f.File.UUID.String(), postsResp.Post.Photo.ID)
+	//as.Equal(f.Meetings[0].UUID.String(), postsResp.Post.Meeting.ID)
```
Baggerone: Should this still be commented out?

application/gqlgen/post.go
```diff
@@ -281,6 +282,14 @@ func convertGqlPostInputToDBPost(ctx context.Context, input postInput, currentUs
 		}
 	}
 
+	if input.EventID != nil {
```
Baggerone: (MeetingID?)

Schparky:
> What the UI exposes to the user can use "event", but the gql schema should still be based on the back end terminology, right?

I guess you're right. I did find it confusing and I got it wrong a few times.

application/actions/post_test.go
Schparky: Yes, until I get your schema update.

PR #165, opened 2019-12-19
create and update meeting

Baggerone:
Also, ordered list of meetings by StartDate


application/actions/meetings_test.go
```diff
@@ -170,3 +173,80 @@ func (as *ActionSuite) Test_RecentMeetingsQuery() {
...
+			description location { description country latitude longitude }
+			startDate endDate moreInfoURL }}`
+
+	var resp meetingResponse
+	as.NoError(as.testGqlQuery(query, user.Nickname, &resp))
+
+	gotMtg := resp.Meeting
+
+	emptyUUID := uuid.UUID{}
+	as.NotEqual(emptyUUID, gotMtg.ID, "don't want empty UUID")
```
Schparky: I think this isn't testing correctly because they are different types. That's why I don't like `NotEqual`.
```suggestion
	as.True(emptyUUID.String() != gotMtg.ID, "don't want empty UUID")
```

application/gqlgen/meeting.go
```diff
@@ -136,3 +140,115 @@ func (r *queryResolver) Meeting(ctx context.Context, id *string) (*models.Meetin
...
+	extras := map[string]interface{}{
+		"user": cUser.UUID,
+	}
+
+	meeting, err := convertGqlMeetingInputToDBMeeting(ctx, input, cUser)
+	if err != nil {
+		return nil, reportError(ctx, err, "UpdateMeeting.ProcessInput", extras)
+	}
+
+	if err := meeting.Update(); err != nil {
```
Schparky: Probably need to check for authorization.

application/gqlgen/meeting.go
```diff
@@ -136,3 +140,115 @@ func (r *queryResolver) Meeting(ctx context.Context, id *string) (*models.Meetin
...
+		return nil, reportError(ctx, err, "CreateMeeting.ProcessInput", extras)
+	}
+
+	location := convertGqlLocationInputToDBLocation(*input.Location)
+	if err = location.Create(); err != nil {
+		return nil, reportError(ctx, err, "CreateMeeting.SetLocation", extras)
+	}
+	meeting.LocationID = location.ID
+
+	if err = meeting.Create(); err != nil {
```
Schparky: Probably need to check for authorization.

fillup: We can put some placeholder authz in there, but I don't think we actually need to restrict who can create them, at least not for now.

application/gqlgen/meeting.go
fillup: True, here we should make sure it is either an org admin for the meeting org, or the creator. (or a super admin)

application/actions/meetings_test.go
Baggerone: Good catch.

application/gqlgen/meeting.go
Baggerone: I added the placeholder

application/gqlgen/meeting.go
Baggerone: Done (although we aren't yet associating orgs with meetings, so I just allow any admins for now)

Schparky: @Baggerone could you look up the post org and only allow admins from that org?

application/models/meeting_fixtures_test.go
```diff
@@ -91,3 +92,37 @@ func createMeetingFixtures_FindByTime(ms *ModelSuite) Meetings {
 	}
 	return meetings
 }
+
+func createMeetingFixtures_CanUpdate(ms *ModelSuite) meetingFixtures {
```
Schparky: Since the UUT doesn't access the database, this test could just create in-memory fixtures. But having said that, I think my other suggestion will require database access.

application/gqlgen/meeting.go
Baggerone: I think we should probably just wait until we associate an org or orgs with the event directly. We just need to decide how that's going to look.

PR #166, opened 2019-12-20
specify file upload error status in same way as other appErrors

Baggerone:
Billy needs this change

### DIFF
--- a/application/actions/upload_test.go
+++ b/application/actions/upload_test.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"net/http"
 	"testing"
 	"time"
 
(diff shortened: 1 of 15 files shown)


application/models/file.go
```diff
@@ -16,6 +16,12 @@ import (
 	"github.com/gofrs/uuid"
 )
 
+type FileUploadError struct {
```
Schparky: I like this implementation. I've never created a custom error type, but I'm curious why this doesn't have a `Error()` method.

application/actions/upload_test.go
```diff
@@ -129,4 +126,46 @@ func (as *ActionSuite) Test_Upload() {
 	as.Regexp("^https?", u.URL)
 	as.Equal("image/gif", u.ContentType)
 	as.Equal(6, u.Size)
+
```
Schparky: Thanks for adding to this test.

application/models/file.go
Baggerone: Funny enough, I wasn't thinking of it as implementing `error`.  I just gave it the attributes I wanted to be able to pull out of it.  I guess I could have given it and `Error()` method that just returns the `Message`. Do you think that's worth doing?

Schparky: Seems like it would make sense. Fewer lines affected maybe.

PR #177, opened 2020-01-03
Before reading file contents, check size from file header

fillup:
I noticed we were checking file size after reading it into a `[]byte` before storing in S3. Since the size is available as part of the `binding.File` I'm adding an earlier check here as well.


application/actions/upload.go
```diff
@@ -37,6 +37,16 @@ func uploadHandler(c buffalo.Context) error {
 		}))
 	}
 
+	if f.Size > int64(domain.MaxFileSize) {
+		domain.ErrLogger.Printf("file upload size (%v) greater than max file size (%v)", f.Size, domain.MaxFileSize)
+		return c.Render(http.StatusInternalServerError, render.JSON(UploadResponse{
```
Baggerone: Should the http status code here match the one two lines down?

fillup: yes, fixed

PR #185, opened 2020-01-07
embed email logo file using multi-part MIME

Schparky:


application/aws/aws.go
```diff
@@ -169,44 +173,98 @@ func SendEmail(to, from, subject, body string) error {
...
+		if err != nil {
+			domain.ErrLogger.Printf("failed to write logo to email, %s", err)
+		}
+		err = encoder.Close()
+		if err != nil {
+			domain.ErrLogger.Printf("failed to close logo base64 encoder, %s", err)
+		}
+	}
+
+	if err = relatedWriter.Close(); err != nil {
```
Baggerone: I'm wondering if it would be better to put these in a `defer` and then maybe you could put a `return` in the `if` block and not have an `else` block.

application/aws/aws.go
```diff
@@ -169,44 +173,98 @@ func SendEmail(to, from, subject, body string) error {
...
-	result, err := svc.SendEmail(input)
+	b := &bytes.Buffer{}
+
+	b.WriteString("From: " + from + "\n")
+	b.WriteString("To: " + to + "\n")
+	b.WriteString("Subject: " + subject + "\n")
+	b.WriteString("MIME-Version: 1.0\n")
+
+	alternativeWriter := multipart.NewWriter(b)
+	b.WriteString(`Content-Type: multipart/alternative; boundary="` + alternativeWriter.Boundary() + `"` + "\n\n")
```
Baggerone: Because of all the + symbols, would it be neater to use fmt.Sprintf here?

Schparky:
> Wow, looks tricky to figure out.

Yeah, I just about left it as a raw text representation rather than using code to generate it. Would be easier to read, but less flexible. If you have ideas for modularizing it, I'm open. Maybe not worth it until other attachments are needed?

PR #188, opened 2020-01-08
Add space to initial nicknames

Baggerone:


application/models/user.go
```diff
@@ -425,7 +425,7 @@ func (u *User) uniquifyNickname() error {
 
 	// User the first nickname prefix that makes it unique
 	for _, p := range allPrefixes() {
-		u.Nickname = p + simpleNN
+		u.Nickname = p + " " + simpleNN
```
Schparky: I believe @forevermatt suggested a space between the first name and last initial as well.

Baggerone: Changed there too

PR #191, opened 2020-01-09
Add GraphQL and data model for post "Watches"

Schparky:
This is a building block to provide notifications to users for posts that may be of interest. Currently only location-based watches are implemented, and the actual notifications for these watches are not sent yet.


application/migrations/20200107221947_create_watches.up.fizz
```diff
@@ -0,0 +1,10 @@
+create_table("watches") {
+	t.Column("id", "integer", {primary: true})
+	t.Column("uuid", "uuid", {})
+	t.Column("owner_id", "integer", {})
+	t.Column("location_id", "integer", {null: true})
+	t.ForeignKey("owner_id", {"users": ["id"]}, {"on_delete": "cascade"})
+	t.Foreign_key("location_id", {"locations": ["id"]}, {"on_delete": "set null"})
```
Baggerone: Don't you want this to be "cascade" too?
If I'm understanding this, the watch starts by being associated with a Location. So, we will end up notifying that users when there is something happening in relation to that Location (e.g. a Post is created near it).

If for some strange reason that Location is deleted, then we would want this watch to be deleted also.
Otherwise, this watch might end up being valid for all locations or something, right?

application/models/watch_test.go
```diff
@@ -0,0 +1,163 @@
...
+		{name: "blank uuid", uuid: "", wantErr: "watch uuid must not be blank"},
+		{name: "wrong uuid", uuid: domain.GetUUID().String(), wantErr: "no rows in result set"},
+	}
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			var watch Watch
+			err := watch.FindByUUID(test.uuid)
+			if test.wantErr != "" {
+				ms.Error(err)
+				ms.Contains(err.Error(), test.wantErr, "wrong error type", err)
```
Baggerone:
```suggestion
				ms.Contains(err.Error(), test.wantErr, "wrong error type")
```

application/migrations/20200107221947_create_watches.up.fizz
Schparky: Yeah, that makes sense.

PR #193, opened 2020-01-10
send notifications for Post Watches

Schparky:
I was able to tap into the existing new request notifications for this, so it's relatively simple.

### DIFF
--- a/application/models/watch.go
+++ b/application/models/watch.go
@@ -10,6 +10,8 @@ import (
 	"github.com/gobuffalo/validate"
 	"github.com/gobuffalo/validate/validators"
 	"github.com/gofrs/uuid"
+
+	"github.com/silinternational/wecarry-api/domain"
 )
 
 // Watch is the model for storing post watches that trigger notifications on the conditions specified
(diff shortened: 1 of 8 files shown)


application/models/watch.go
```diff
@@ -113,3 +115,20 @@ func (w *Watch) SetLocation(location Location) error {
 func (w *Watch) Destroy() error {
 	return DB.Destroy(w)
 }
+
+// matchesPost returns true if the Watch matches the Post
```
Baggerone:
```suggestion
// matchesPost returns true if the Watch's Location is near the Post's Destination
```

application/models/watch_test.go
```diff
@@ -161,3 +161,39 @@ func (ms *ModelSuite) TestWatch_GetSetLocation() {
...
 	ms.InDelta(newLoc.Longitude.Float64, got.Longitude.Float64, 0.0001, "longitude doesn't match")
 }
+
+func (ms *ModelSuite) TestWatch_matchesPost() {
+	posts := createPostFixtures(ms.DB, 1, 0, false)
+	watches := createWatchFixtures(ms.DB, createUserFixtures(ms.DB, 1).Users)
+
+	dest, err := posts[0].GetDestination()
+	ms.NoError(err)
+	ms.NoError(watches[0].SetLocation(*dest))
```
Baggerone: Watches will have different Location instances than posts, right? So, would it be better to create a separate Location just for the Watch that has similar values as the Post's Location?

application/templates/mail/new_request.plush.html
```diff
@@ -1,4 +1,11 @@
+<h4>New Request</h4>
 <p>
-    There is a new request on <a href="<%= uiURL %>"><%= appName %></a>:
+    <a href="<%= postURL %>"><%= postTitle %></a>
```
Baggerone: The other messages have this as an `<h4>`. Is it worth being consistent here?

application/templates/mail/new_request.plush.html
```diff
@@ -1,4 +1,11 @@
...
+</p>
+<p>
+    There is a new request on <a href="<%= uiURL %>"><%= appName %></a> that we thought you might be able to help with.
```
Baggerone:
```suggestion
    There is a new request on <a href="<%= uiURL %>"><%= appName %></a> that we thought might be of interest to you.
```

Baggerone: (Trying to avoid having them feel pressured)

application/templates/mail/new_request.plush.html
Schparky: Probably. Would it be helpful to have the request description here, or do we want to entice the reader into the app?

Baggerone: @fillup What do you think - include the new post's description?

PR #195, opened 2020-01-10
For Discussion only - multiple committers for a post

Baggerone:
This is my first quick and dirty attempt at working through changes for allowing multiple committers on a Post.  I didn't do anything to fix current tests or create new ones.

Is it worth the added complexity in the code and in our future brains.


application/gqlgen/schema.graphql
```diff
@@ -312,6 +312,7 @@ input LocationInput {
 input UpdatePostStatusInput {
     id: ID!
     status: PostStatus!
+    userID: ID
```
Schparky: I wonder if the COMMITTED to ACCEPTED transition would be better as a new mutation since it's the only one that needs a User ID.

Baggerone: Phillip was wondering about removing the possibility of a committed status.  That would simplify the code a bit and then make it necessary to add a new mutation with the new potential provider.
Then also, all the Posts we currently consider COMMITTED would end up being OPEN, allowing request for a list of Posts to stay what it is now.

Schparky: So that would do away with the whole idea of "committed" and just leave that process up to user-to-user communication. What if the requester just chooses from the users that have an open thread? No new tables and only one new mutation to select the provider.

Baggerone: I think we still need to keep track of potential providers in the API.  I doubt if the UI guys would want to do that.

Schparky: I wasn't suggesting that the UI kept track of potential providers. My thought was that merely opening a thread qualifies a user as a "potential provider". That only means that the requester would see an "accept" button on that thread. It would be up to the human interaction to determine whether that makes sense for any given user.

Baggerone: One reason for the API to keep track of the explicit potential providers is so it can notify the rejected ones once the requester has accepted a different offer.

Schparky: I hadn't thought of that. Still, that could be either implied by lack of communication or left up to the requester to let others know it was handled. Not ideal, but simple.

PR #196, opened 2020-01-10
strip EXIF metadata from images

Schparky:


application/models/file.go
```diff
@@ -92,6 +99,17 @@ func (f *File) Store(name string, content []byte) *FileUploadError {
 		return &e
 	}
 
+	// if possible, re-encode the image to strip EXIF metadata
+	img, _, err := image.Decode(bytes.NewReader(content))
+	if err == nil {
+		buf := new(bytes.Buffer)
+		if err := jpeg.Encode(buf, img, nil); err == nil {
```
fillup: Do I understand this correct that we'll decode gif, jpg, png, and webp, and then just encode them all to jpg, essentially converting some file types? That seems fine to me, just making sure I understand.

application/models/file.go
```diff
@@ -92,6 +99,17 @@ func (f *File) Store(name string, content []byte) *FileUploadError {
...
+			content = buf.Bytes()
+			contentType = "image/jpg"
+			name = name + ".jpeg"
```
fillup: Does the `name` being sent into this function already include the file extension? Will this result in `file.jpg.jpeg` and `file.png.jpeg`?

application/models/file.go
Schparky: Yes, that's correct. It was a lazy approach to not bother re-encoding back to the original format. Also, there is no webp encoder in the standard library.

application/models/file.go
Schparky: Yes, also a lazy approach. It has the "benefit" of seeing the original format. Really won't be visible anywhere except S3 and possibly if a user asked the browser to "save image". Maybe better to just rename the file to `id + ".jpeg"`?

PR #199, opened 2020-01-13
Scrutinizer configured for static analysis and code coverage

fillup:


application/database.yml
```diff
@@ -6,11 +6,18 @@ development:
   host: db
   pool: 5
 
-test:
-  url: {{envOr "TEST_DATABASE_URL" "postgres://wecarry:wecarry@testdb:5432/wecarry_test?sslmode=disable"}}
+#test:
+#  url: {{envOr "TEST_DATABASE_URL" "postgres://wecarry:wecarry@testdb:5432/wecarry_test?sslmode=disable"}}
```
Schparky: Can this be put back as it was? Does it affect local testing?

fillup: it has to use the new database and username to work with scrutinizer, so those lines can actually be deleted. I'll take them out

PR #203, opened 2020-01-16
validate auth email against the email back from the auth provider

Baggerone:


application/actions/auth.go
```diff
@@ -315,6 +322,24 @@ func authCallback(c buffalo.Context) error {
...
+	respDomain := domain.EmailDomain(authRespEmail)
+
+	if emailDomain == respDomain {
+		domain.Warn(c, "authentication emails don't match: "+originalAuthEmail+
+			" vs. "+authRespEmail)
+		return nil
+	}
+
+	return errors.New("authentication email domains don't match: " + originalAuthEmail +
+		" vs. " + authRespEmail)
```
Schparky: Inconsistent spacing around `+`. Feel free to use `Sprintf` if you prefer that. I prefer `+` but don't want to dictate.

Baggerone: Oh well. My go formatter is doing that.  I take out the spaces on the one, and it adds them back in.  I add the spaces on the other, and it takes them out.

application/actions/auth.go
```diff
@@ -315,6 +322,24 @@ func authCallback(c buffalo.Context) error {
...
+func verifyEmails(c buffalo.Context, originalAuthEmail, authRespEmail string) error {
+	if originalAuthEmail == authRespEmail {
+		return nil
+	}
+
+	emailDomain := domain.EmailDomain(originalAuthEmail)
+	respDomain := domain.EmailDomain(authRespEmail)
+
+	if emailDomain == respDomain {
+		domain.Warn(c, "authentication emails don't match: "+originalAuthEmail+
```
Baggerone:
```suggestion
		domain.Warn(c, "authentication emails don't match: " + originalAuthEmail +
```

application/actions/auth.go
```diff
@@ -315,6 +322,24 @@ func authCallback(c buffalo.Context) error {
...
+			" vs. "+authRespEmail)
```
Baggerone:
```suggestion
			" vs. " + authRespEmail)
```

application/actions/auth.go
Schparky: That's weird!

Schparky: Looks like a bug in `go fmt` to me. I played with it and it seems that  function calls with more than one parameter do not allow spaces around `+` but function calls with only one parameter require a space around `+`. Maybe let it do its thing or extract the string to a local variable so the parameter is just a variable. Otherwise, the next time `go fmt` runs an this file, it will modify these lines again.

PR #204, opened 2020-01-16
Allow for trusted (connected) organizations

Schparky:
Implemented with a new `organization_trusts` table with two `organizations` foreign key columns. When an `OrganizationTrust` is created, two records are created to make queries slightly more simple and provide future option for asymmetric relationships.


application/models/trust.go
```diff
@@ -0,0 +1,104 @@
...
+func (t *Trust) ValidateUpdate(tx *pop.Connection) (*validate.Errors, error) {
+	return validate.NewErrors(), nil
+}
+
+// Create stores the Trust data as a new record in the database. Also creates a mirrored copy of the given Trust.
+func (t *Trust) Create() error {
+	var t2 Trust
+	if err := t2.FindByOrgIDs(t.PrimaryID, t.SecondaryID); err == nil {
+		// already exists
+		return nil
```
Baggerone: In this case, I wonder if it should still ensure that the mirror exists.

application/models/trust.go
```diff
@@ -0,0 +1,104 @@
...
+		return err
+	}
+	return nil
+}
+
+// Remove destroys two Trust records identified by the given Organization IDs
+func (t *Trust) Remove(orgID1, orgID2 int) error {
+	var t1, t2 Trust
+	if err := t1.FindByOrgIDs(orgID1, orgID2); err != nil {
+		return fmt.Errorf("remove Trust failed locating the first Trust record, %s", err)
```
Baggerone: In this case, wouldn't we still want to try to remove the mirror?

application/gqlgen/mutations.go
```diff
@@ -162,3 +162,51 @@ func (r *mutationResolver) SetThreadLastViewedAt(ctx context.Context, input SetT
...
+	extras := map[string]interface{}{
+		"user": cUser.UUID,
+	}
+
+	var organization models.Organization
+	if err := organization.FindByUUID(input.PrimaryID); err != nil {
+		return nil, reportError(ctx, err, "CreateTrust.FindPrimaryOrganization", extras)
+	}
+
+	if !cUser.CanEditOrganization(organization.ID) {
```
Baggerone: Maybe use a new CanCreateTrust function that is more strict.

application/locales/global.en.yaml
```diff
@@ -244,3 +244,17 @@
...
 - id: RemoveWatch.FindByUser
   translation: We had a problem getting a list of remaining Watches
+
+# Trust
+- id: GetOrganizationTrustedOrganizations
+  translation: We had a problem getting a list of trusted organizations
+- id: CreateTrust.FindPrimaryOrganization
+  translation: We had a problem finding the organization to create the Trust
+- id: CreateTrust.Unauthorized
+  translation: We had a problem finding the organization to create the Trust
```
Baggerone: Missing just plain `CreateTrust`?

application/actions/organization_fixtures_test.go
```diff
@@ -238,14 +239,57 @@ func fixturesForOrganizationDomain(as *ActionSuite) OrganizationFixtures {
...
+	orgs[1].Name = "trusted org 1"
+	orgs[2].Name = "trusted org 2"
+	for i := range orgs {
+		orgs[i].AuthType = models.AuthTypeSaml
+		orgs[i].AuthConfig = "{}"
+		orgs[i].Url = nulls.NewString("https://www.example.com")
+		test.MustCreate(as.DB, &orgs[i])
+	}
+
+	trust := models.Trust{PrimaryID: orgs[0].ID, SecondaryID: orgs[1].ID}
```
Baggerone: The mirror trust should get created too, right? (If so, I guess your desired test results may change.)

application/actions/organization_fixtures_test.go
```diff
@@ -258,7 +302,42 @@ func fixturesForUpdateOrganization(as *ActionSuite) OrganizationFixtures {
...
+	orgs[1].Name = "trusted org 1"
+	orgs[2].Name = "trusted org 2"
+	for i := range orgs {
+		orgs[i].AuthType = models.AuthTypeSaml
+		orgs[i].AuthConfig = "{}"
+		orgs[i].Url = nulls.NewString("https://www.example.com")
+		test.MustCreate(as.DB, &orgs[i])
+	}
+
+	trusts := models.Trusts{
```
Baggerone: Create the mirrors also?

application/models/trust.go
Schparky: Yes, good idea.

application/models/trust.go
Schparky: Yes.

application/gqlgen/mutations.go
Schparky: Sounds good.

application/locales/global.en.yaml
Schparky: Right.

application/actions/organization_fixtures_test.go
Schparky: This creates both.

application/actions/organization_fixtures_test.go
Schparky: It does create both.

application/models/organizationtrust.go
```diff
@@ -0,0 +1,121 @@
...
+	if err := o.Create(); err != nil {
+		return err
+	}
+
+	mirror := OrganizationTrust{
+		PrimaryID:   o.SecondaryID,
+		SecondaryID: o.PrimaryID,
+	}
+	if err := mirror.Create(); err != nil {
+		_ = DB.Destroy(o)
```
Baggerone: Maybe return the error from here?

application/models/organizationtrust.go
```diff
@@ -0,0 +1,121 @@
...
+	return nil
+}
+
+// Remove destroys the OrganizationTrust record identified by the given Organization IDs
+func (o *OrganizationTrust) Remove(orgID1, orgID2 int) error {
+	var trust OrganizationTrust
+	if err := trust.FindByOrgIDs(orgID1, orgID2); err != nil {
+		if domain.IsOtherThanNoRows(err) {
+			return err
+		} else {
```
Baggerone: with the `return err` above this `else` isn't needed (just its contents)

application/gqlgen/mutations.go
```diff
@@ -162,3 +162,51 @@ func (r *mutationResolver) SetThreadLastViewedAt(ctx context.Context, input SetT
...
+	extras := map[string]interface{}{
+		"user": cUser.UUID,
+	}
+
+	var organization models.Organization
+	if err := organization.FindByUUID(input.PrimaryID); err != nil {
+		return nil, reportError(ctx, err, "RemoveOrganizationTrust.FindPrimaryOrganization", extras)
+	}
+
+	if !cUser.CanCreateOrganizationTrust() {
```
Baggerone: I'm thinking simple Org Admins should be able to remove an OrgTrust for their own org. @fillup, what do you think?

fillup: Sure, seems fair. But I'm okay to leave it this way for now too since we wont need that ability for a while.

PR #206, opened 2020-01-21
delete unlinked files

Schparky:


application/models/file.go
```diff
@@ -206,3 +206,76 @@ func (f *File) Create() error {
...
 }
+
+// DeleteUnlinked removes all files that are no longer linked to any database records
+func (f *Files) DeleteUnlinked() error {
+	var files Files
+	if err := DB.Select("id").All(&files); err != nil {
+		return err
+	}
+
+	toDelete := make(map[int]bool, len(files))
```
Baggerone: Does using the len() for creating a map do anything useful.

application/models/file.go
```diff
@@ -206,3 +206,76 @@ func (f *File) Create() error {
...
+	var users Users
+	if err := DB.Select("photo_file_id").Where("photo_file_id is not null").All(&users); err != nil {
+		return err
+	}
+	for _, u := range users {
+		toDelete[u.PhotoFileID.Int] = false
+	}
+
+	for id, del := range toDelete {
+		if del {
```
Baggerone: How about `if !del { continue }`

application/models/file.go
Schparky: I had the same question, and I think it allocates enough memory for all of the pointers.

PR #209, opened 2020-01-22
filter nicknames

Schparky:


application/domain/domain.go
```diff
@@ -524,3 +525,39 @@ func EmailFromAddress(name *string) string {
 	}
 	return addr
 }
+
+// RemoveUnwantedChars removes characters from `str` that are not in `allowed` and not in "safe" character ranges.
+func RemoveUnwantedChars(str, allowed string) string {
+	return strings.Map(func(r rune) rune {
```
Baggerone: I think this would be more readable to me ...
```
	removeFunc := func(r rune) rune {
		if strings.IndexRune(allowed, r) >= 0 || isSafeRune(r) {
			return r
		}
		return -1
	}
	return strings.Map(removeFunc, str)
```

application/models/user.go
```diff
@@ -434,6 +434,7 @@ func (u *User) GetPhotoURL() (*string, error) {
 
 // Save wraps DB.Save() call to check for errors and operate on attached object
 func (u *User) Save() error {
+	u.Nickname = domain.RemoveUnwantedChars(u.Nickname, "-_ .,'&@")
```
Baggerone: Should we ensure the resulting string has at least one visible character in it? (Would invisible characters just be spaces, and zero-width joiners?)

Schparky: Good thought. I just checked and it is handled by the User validation. The UI gives the "Something went wrong We had a problem updating the user profile." message and "nickname: Nickname can not be blank." is logged.

application/domain/domain.go
Schparky: That is better.

application/models/user.go
Schparky: I somehow ignored the word "visible" in your message. I'll take another look at this.

Schparky: Done

PR #216, opened 2020-01-30
add ability for org domains to override auth config

fillup:


application/actions/auth.go
```diff
@@ -222,7 +222,7 @@ func authRequest(c buffalo.Context) error {
 	orgID := org.UUID.String()
 	c.Session().Set(OrgIDSessionKey, orgID)
 
-	sp, err := org.GetAuthProvider()
+	sp, err := org.GetAuthProvider(authEmail)
```
Baggerone: The param is also needed on lines 272 and 402. I'm thinking that for line 402, you will need to first get the authEmail - maybe the way it's done in `authCallback()`

application/models/organization_test.go
```diff
@@ -644,3 +646,51 @@ func (ms *ModelSuite) TestOrganization_AttachLogo() {
...
+	ms.NoError(err, "unable to create orgDomain2 fixture")
+
+	var o Organization
+	err = o.FindByUUID(uid.String())
+	ms.NoError(err, "unable to find organization fixture")
+
+	// should get type facebook:
+	provider, err := o.GetAuthProvider("test@domain1.com")
+	ms.NoError(err, "unable to get authprovider for test@domain1.com")
+	if _, ok := provider.(*facebook.Provider); !ok {
```
Baggerone: Looks like a shorter options is ...
```
	ms.IsType(&facebook.Provider{}, provider, "auth provider not expected facebook type")
```
(Same for next one)

application/migrations/20200130145100_add_org_domain_authconfig.up.fizz
```diff
@@ -0,0 +1,2 @@
+add_column("organization_domains", "auth_type", "string", {null: true})
```
Schparky: Do these need to be nullable or just default to an empty string? Your check in `GetAuthProvider` is against an empty string (not `AuthType.Valid`) anyway.

application/models/organizationdomain.go
```diff
@@ -48,3 +51,12 @@ func (o *OrganizationDomain) GetOrganizationUUID() (string, error) {
 func (o *OrganizationDomain) Create() error {
 	return create(o)
 }
+
+func (o *OrganizationDomain) FindByDomain(domainName string) error {
```
Schparky: to satisfy golint
```suggestion
// FindByDomain finds a record by the domain name
func (o *OrganizationDomain) FindByDomain(domainName string) error {
```

application/models/organization.go
```diff
@@ -70,21 +70,35 @@ func (o *Organization) ValidateUpdate(tx *pop.Connection) (*validate.Errors, err
 	return validate.NewErrors(), nil
 }
 
-func (o *Organization) GetAuthProvider() (auth.Provider, error) {
+func (o *Organization) GetAuthProvider(authEmail string) (auth.Provider, error) {
```
Schparky: to satisfy go lint
```suggestion
// GetAuthProvider returns the auth provider associated with the domain of `authEmail`, if assigned, otherwise from the Organization's auth provider.
func (o *Organization) GetAuthProvider(authEmail string) (auth.Provider, error) {
```

application/migrations/20200130145100_add_org_domain_authconfig.up.fizz
fillup: Interesting, good point, I'm not really sure. I assume most of the time these fields wont be used, so I just went with null, but if the other is better feel free to change it.

application/gqlgen/mutations.go
```diff
@@ -110,6 +111,43 @@ func (r *mutationResolver) CreateOrganizationDomain(ctx context.Context, input C
...
+		err := errors.New("insufficient permissions")
+		return nil, reportError(ctx, err, "UpdateOrganizationDomain.Unauthorized", extras)
+	}
+
+	var orgDomain models.OrganizationDomain
+	if err := orgDomain.FindByDomain(input.Domain); err != nil {
+		return nil, reportError(ctx, err, "UpdateOrganizationDomain.NotFound", extras)
+	}
+
+	orgDomain.AuthType = *input.AuthType
```
Schparky: I'm not sure if this is needed, but you did it above, so may as well be consistent.
```suggestion
	orgDomain.AuthType = domain.ConvertStrPtrToString(input.AuthType)
```

application/gqlgen/mutations.go
```diff
@@ -110,6 +111,43 @@ func (r *mutationResolver) CreateOrganizationDomain(ctx context.Context, input C
...
+	orgDomain.AuthConfig = *input.AuthConfig
```
Schparky:
```suggestion
	orgDomain.AuthConfig = domain.ConvertStrPtrToString(input.AuthConfig)
```

application/actions/organizationdomain_test.go
```diff
@@ -43,4 +43,35 @@ func (as *ActionSuite) Test_CreateOrganizationDomain() {
...
+	// User with admin role but invalid domain - should error
+	inValidUpdateInput := fmt.Sprintf(`organizationID:"%s" domain:"%s" authType:"saml", authConfig:"{}"`,
+		f.Organizations[0].UUID.String(), "invalid.com")
+
+	query = fmt.Sprintf("mutation{domain: updateOrganizationDomain(input: {%s}) {%s}}",
+		inValidUpdateInput, allFieldsQuery)
+	var resp4 OrganizationDomainResponse
+	err = as.testGqlQuery(query, f.Users[1].Nickname, &resp4)
+	as.Error(err)
+
```
Baggerone: Want to add a test for nulling-out the auth fields?

application/gqlgen/mutations.go
Baggerone: I think these panic if the fields are null

application/gqlgen/schema.graphql
```diff
@@ -207,11 +208,15 @@ input UpdateOrganizationInput {
 type OrganizationDomain {
     domain: String!
     organizationID: ID!
+    authType: String!
+    authConfig: String!
 }
 
 input CreateOrganizationDomainInput {
```
Baggerone: Should we have a separate UpdateOrganizationDomainInput that allows the user to modify the domain? I guess we would need two domain fields - one to find the current object and a second for the new domain value.

fillup: i don't think we need to allow them to update the domain name itself, they can just recreate it.

PR #217, opened 2020-02-03
make status check touch db

Baggerone:


Baggerone:
If I kill my db locally, the old code still reports the status as "good".

I'll look into doing a `First` style query.

PR #220, opened 2020-02-04
Feature/needed before date

Baggerone:


application/actions/post_test.go
```diff
@@ -112,6 +115,10 @@ func (as *ActionSuite) Test_PostQuery() {
 	as.Equal(f.Posts[0].Title, resp.Post.Title)
 	as.Equal(f.Posts[0].Description.String, resp.Post.Description)
 
+	wantDate := f.Posts[0].NeededBefore.Time
+	wantDate = time.Date(wantDate.Year(), wantDate.Month(), wantDate.Day(), 0, 0, 0, 0, time.UTC)
```
Schparky: Is this right? Looks the the first assignment is not used.

application/actions/post_test.go
```diff
@@ -236,17 +245,19 @@ func (as *ActionSuite) Test_CreatePost() {
 			visibility: ALL
 		`
 	query := `mutation { post: createPost(input: {` + input + `}) { organization { id } photo { id } type title
-			description destination { description country latitude longitude }
+			neededBefore description destination { description country latitude longitude }
 			origin { description country latitude longitude }
 			size url kilograms visibility }}`
 
 	as.NoError(as.testGqlQuery(query, f.Users[0].Nickname, &postsResp))
+	//var nilTime *time.Time
```
Schparky:
```suggestion
```

application/gqlgen/gqlgen.yml
```diff
@@ -47,6 +47,8 @@ models:
     fields:
       id:
         resolver: true
+      neededBefore:
+        resolver: true
```
Schparky: I'm just curious. Did you try it without a custom resolver?

application/gqlgen/post.go
```diff
@@ -263,6 +273,10 @@ func convertGqlPostInputToDBPost(ctx context.Context, input postInput, currentUs
 
 	setOptionalStringField(input.Title, &post.Title)
 
+	if input.NeededBefore != nil {
+		post.NeededBefore = nulls.NewTime(*input.NeededBefore)
+	}
+
```
Schparky: Do we need a way to "erase" an existing needed-before time?

application/actions/post_test.go
Baggerone: It's dropping off the hours-minutes... from the fixture time.

Schparky: Oh, I see it now. Sorry, read too quickly.

application/gqlgen/gqlgen.yml
Baggerone: If I don't do that, I get this ...
```
gqlgen/post.go:17:23: cannot use &postResolver literal (type *postResolver) as type PostResolver in return argument:
	*postResolver does not implement PostResolver (missing NeededBefore method)
```
Is there something else I need to add/remove, do you think?

Schparky: Looks like its logic automatically indicates that a resolver is needed. To get that error, did you also remove your `NeededBefore` resolver?

application/models/models.go
```diff
@@ -92,6 +92,17 @@ func GetIntFromNullsInt(in nulls.Int) *int {
 	return &output
 }
 
+// GetTimeFromNullsTime returns a pointer to a string that looks
```
Schparky:
```suggestion
// GetStringFromNullsTime returns a pointer to a string that looks
```

application/models/post.go
```diff
@@ -363,6 +376,13 @@ func (p *Post) ValidateCreate(tx *pop.Connection) (*validate.Errors, error) {
 	//return validate.NewErrors(), nil
 }
 
+type futureDateValidator struct {
```
Schparky: Is this still needed?

application/models/post.go
```diff
@@ -325,6 +325,16 @@ func (p *Post) SetProviderWithStatus(status PostStatus, currentUser User) {
 
 // Validate gets run every time you call a "pop.Validate*" (pop.ValidateAndSave, pop.ValidateAndCreate, pop.ValidateAndUpdate) method.
 func (p *Post) Validate(tx *pop.Connection) (*validate.Errors, error) {
+	tomorrow := time.Now()
```
Schparky: Will this work:
```
tomorrow := time.Now().Truncate(domain.DurationDay).Add(domain.DurationDay)
```

application/models/post.go
Baggerone: Oh yeah. I found the built in validator after I started on that.

application/models/post.go
Baggerone: That looks a lot better.

PR #222, opened 2020-02-04
add test cases to TestPosts_FindByUser

Schparky:


application/models/post_fixtures_test.go
```diff
@@ -196,16 +196,25 @@ func CreateFixtures_Posts_FindByUser(ms *ModelSuite) PostFixtures {
...
+	ms.NoError(DB.UpdateColumns(&uo, "organization_id"))
+
+	uo, err = users[3].FindUserOrganization(orgs[0])
+	ms.NoError(err)
+	uo.OrganizationID = orgs[1].ID
+	ms.NoError(DB.UpdateColumns(&uo, "organization_id"))
+
 	posts := createPostFixtures(ms.DB, 8, 0, false)
 	posts[1].OrganizationID = orgs[1].ID
 	posts[2].Status = PostStatusCommitted
```
Baggerone: Maybe choose a status that's going to still be in the code, e.g. `PostStatusAccepted`?

application/models/post_fixtures_test.go
```diff
@@ -177,15 +177,15 @@ func createFixturesForPostFindByUserAndUUID(ms *ModelSuite) PostFixtures {
...
+//        Org0                Org1           Org2
+//        |  |                | | |          | |
+//        |  +----+-----------+ | +----+-----+ +
+//        |       |             |      |       |
+//       User1  User0        User3   Trust   User2
 //
-// Org0: Post0, Post2, Post3, Post4
-// Org1: Post1
-// Org2: Post5, Post6, Post7
+// Org0: Post0 (SAME), Post2 (SAME), Post3 (SAME), Post4 (SAME)
```
Baggerone: It might be worth including the post status in the parenthesis or something that indicates whether it should be filtered out of the list.

application/models/post_fixtures_test.go
Schparky: Can't go from OPEN (as created by createPostFixtures) to ACCEPTED in one step, so this code would have to be edited again to remove the double-update when dev2 is merged. Hopefully this won't be a merge conflict since line 220 is separated from the new lines.

application/models/post_fixtures_test.go
Schparky: Good suggestion.

PR #226, opened 2020-02-06
Support removal of photo or location from user profile

Schparky:
This is the first of the updates required to implement the recent decision to use `null` in GraphQL to remove data from the respective object.


application/gqlgen/user.go
```diff
@@ -202,7 +209,7 @@ func (r *mutationResolver) UpdateUser(ctx context.Context, input UpdateUserInput
 		}
 	}
 
-	if err := user.Save(); err != nil {
+	if err = user.Save(); err != nil {
```
Baggerone: I don't think it hurt that you removed the colon, but I also don't think it was necessary (because of the `if`).

application/models/user.go
```diff
@@ -516,6 +538,18 @@ func (u *User) SetLocation(location Location) error {
 	return u.Save()
 }
 
+// RemoveLocation removes the location record associated with the user
+func (u *User) RemoveLocation() error {
+	if u.LocationID.Valid {
```
Baggerone: Unnecessarily nested `if`, but the function is so short that I don't mind if you leave it this way. (-;

application/models/user.go
```diff
@@ -428,6 +428,28 @@ func (u *User) AttachPhoto(fileID string) (File, error) {
 	return f, nil
 }
 
+// RemovePhoto removes an attached photo from the User profile
+func (u *User) RemovePhoto() error {
+	if u.ID < 1 {
+		return fmt.Errorf("invalid User ID %d", u.ID)
+	}
+
+	oldID := u.PhotoFileID
```
Baggerone: Would it be better to have, 
```
if !oldID.Valid {
     return nil
}
```
and then do the UpdateColumns and unlink the file.

application/actions/user_test.go
```diff
@@ -225,19 +225,39 @@ func (as *ActionSuite) TestUpdateUser() {
...
-			ExpectError: true,
+			ExpectError: "not allowed",
+		},
+		{
+			Name: "remove photo",
+			Payload: fmt.Sprintf(`mutation {user: updateUser(input:{id: "%v", location: %v}) {avatarURL}}`,
+				f.Users[1].UUID, location),
+			TestUser: f.Users[0],
+			Test: func(t *testing.T) {
+				as.Equal(f.Users[1].AuthPhotoURL.String, resp.User.AvatarURL, "expected photo to be deleted")
```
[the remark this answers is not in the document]
Schparky: It's a bit misleading because it's a tiered response. If no attached photo, then it uses the auth photo, if no auth photo, then gravatar. Should never be "".

application/gqlgen/schema.graphql
```diff
@@ -70,12 +70,13 @@ enum PostSize {
     XLARGE
 }
 
-"""
-Visibility for Posts, ALL organizations, TRUSTED organizations, or SAME organization only
-"""
+"Visibility for Posts, ALL organizations, TRUSTED organizations, or SAME organization only"
```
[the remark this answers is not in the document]
Schparky: Yes. Both formats are acceptable. They show up in Insomnia's schema browser.

application/gqlgen/schema.graphql
```diff
@@ -70,12 +70,13 @@ enum PostSize {
...
 enum PostVisibility {
+    "Visible to all users from all organizations in the system"
     ALL
+    "Visible to users from all organizations trusted by the viewing user's organization"
```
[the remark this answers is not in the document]
Schparky: Good catch, thanks.

application/models/user.go
Schparky: Probably. As you've probably noticed, I naturally think this way. Maybe a holdover from old C-language tradition?

application/gqlgen/user.go
Schparky: Just silencing a lint warning.

PR #228, opened 2020-02-07
Change domain.ErrLogger to proxy to stderr and Rollbar

fillup:


application/domain/domain.go
```diff
@@ -224,6 +224,34 @@ type AppError struct {
...
+// ErrLogProxy wraps standard error logger plus sends to Rollbar
+type ErrLogProxy struct {
+	LocalLog  log.Logger
+	RemoteLog *rollbar.Client
+}
+
+func (e *ErrLogProxy) SetOutput(w io.Writer) {
+	e.LocalLog.SetOutput(w)
+}
+func (e *ErrLogProxy) Printf(format string, a ...interface{}) {
```
Baggerone: Maybe, add blank line before functions.

application/listeners/posthelpers.go
```diff
@@ -260,7 +260,7 @@ func sendNotificationRequestFromCompletedToAcceptedOrDelivered(params senderPara
 }
 
 func sendNotificationEmpty(params senderParams) {
-	domain.ErrLogger.Print("Notification not implemented yet for " + params.template)
+	domain.ErrLogger.Printf("Notification not implemented yet for %s", params.template)
```
Baggerone: Minor note - you could have left this the same other than adding the `f` to `Print`.  (Same for some of the others.)

Schparky: Could they have been left alone if a `Print` is defined in `ErrLogProxy`?

application/domain/domain.go
fillup: k

PR #233, opened 2020-02-10
make posts.kilograms nullable

Schparky:


application/gqlgen/post.go
```diff
@@ -159,11 +159,13 @@ func (r *postResolver) URL(ctx context.Context, obj *models.Post) (*string, erro
 // Kilograms resolves the `kilograms` property of the post query, converting float64 to string
```
Baggerone:
```suggestion
// Kilograms resolves the `kilograms` property of the post query as a pointer to a float64
```

Schparky: Thanks. I found that the first time through, but missed it on this branch.

PR #239, opened 2020-02-13
Add CompletedOn field to Posts - set when entering Completed status

Baggerone:


application/models/post.go
```diff
@@ -511,6 +511,34 @@ func (p *Post) manageStatusTransition() error {
...
 	emitEvent(e)
 
+	// If completed, hydrate CompletedOn. If not completed, nullify CompletedOn
+	// Don't use p.UpdateColumns, due to this being called by the AfterUpdate function
+	switch p.Status {
+	case PostStatusCompleted:
+		if !p.CompletedOn.Valid {
+			err := DB.RawQuery(
+				fmt.Sprintf(`UPDATE posts set completed_on = '%s' where ID = %v`,
+					time.Now().Format(domain.DateFormat), p.ID)).Exec()
```
Schparky: I believe Pop takes care of the format from a `Time` even in a RawQuery.
```suggestion
					time.Now(), p.ID)).Exec()
```

application/models/post_fixtures_test.go
```diff
@@ -86,12 +87,27 @@ func createFixturesForTestPost_manageStatusTransition_forwardProgression(ms *Mod
...
+		Status: PostStatusAccepted,
+		PostID: posts[3].ID,
+	}
+	mustCreate(ms.DB, &pHistory)
+
+	// Give these new statuses while by-passing the status transition validation
+	for i, status := range [2]PostStatus{PostStatusAccepted, PostStatusDelivered} {
+		id := i + 2
+
+		posts[id].Status = status
```
Schparky: Since actual ID is not the same as the `posts` index, using `id` may be misleading.

application/models/post_test.go
```diff
@@ -956,6 +987,13 @@ func (ms *ModelSuite) TestPost_manageStatusTransition_backwardProgression() {
 			ms.Equal(test.newStatus, ph.Status, "incorrect Status ")
 			ms.Equal(test.post.ReceiverID, ph.ReceiverID, "incorrect ReceiverID ")
 			ms.Equal(test.providerID, ph.ProviderID, "incorrect ProviderID ")
+
+			if test.wantCompletedOn {
+				ms.True(test.post.CompletedOn.Valid, "expected a valid CompletedOn date")
+			} else {
+				ms.False(test.post.CompletedOn.Valid, "expected a null CompletedOn date")
+			}
```
Schparky: `ms.Equal(test.wantCompletedOn, test.post.CompletedOn.Valid, "incorrect CompletedOn value")`

application/models/post_fixtures_test.go
Baggerone: :+1:

application/models/post.go
Baggerone: That change breaks my tests.
```unable to set Post.CompletedOn for ID: 29, pq: invalid input syntax for type date: "2020-02-14 13:55:40.492484744 +0000 UTC m=+29.406122424"```

application/models/post_test.go
Baggerone: :+1:

application/models/post.go
Schparky: Sorry, I Guess I remembered wrong. Maybe it was on a SELECT, though I would be surprised if that was different.

PR #241, opened 2020-02-14
posts query filters

Schparky:


application/models/post.go
```diff
@@ -811,28 +810,37 @@ func (p *Posts) FindByUser(ctx context.Context, user User) error {
...
+	AND status not in (?, ?)`
+
+	args := []interface{}{user.ID, PostVisibilityAll, PostVisibilityTrusted, PostStatusRemoved,
+		PostStatusCompleted}
+
+	if searchText != nil {
+		selectClause = selectClause + " AND (LOWER(title) LIKE ? or LOWER(description) LIKE ?)"
+		likeText := "%" + strings.ToLower(*searchText) + "%"
+		args = append(args, likeText)
+		args = append(args, likeText)
```
Baggerone: I think you can just include this second `likeText` in the first append.

application/models/post.go
```diff
@@ -811,28 +810,37 @@ func (p *Posts) FindByUser(ctx context.Context, user User) error {
...
+	}
 
-	return DB.
-		Scope(scopeUserOrgs(user)).
-		Scope(scopeNotCompleted()).
-		Where(where, pType, contains, contains).
-		Order("created_at desc").
-		All(p)
+	for i := range posts {
+		*p = append(*p, posts[i])
```
Baggerone: It looks like this assumes that p begins as an empty slice. Should we error if it's not?

application/models/post.go
```diff
@@ -1007,3 +1015,28 @@ func (p *Post) Meeting() (*Meeting, error) {
 
 	return &meeting, nil
 }
+
+// FilterDestination returns a list of all posts with a Destination near the given location. The database is not
+// touched.
+func (p Posts) FilterDestination(location Location) Posts {
+	filtered := make(Posts, 0)
+	for i := range p {
+		_ = DB.Load(&p[i], "Destination")
```
Baggerone: Can this go before the `for loop` and act on all the posts at once? (Same for next function)

application/gqlgen/meeting.go
```diff
@@ -236,7 +236,7 @@ func (r *mutationResolver) CreateMeeting(ctx context.Context, input meetingInput
 		return nil, reportError(ctx, err, "CreateMeeting.Unauthorized", extras)
 	}
 
-	location := convertGqlLocationInputToDBLocation(*input.Location)
+	location := convertLocation(*input.Location)
```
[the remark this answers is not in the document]
Schparky: It's an unfortunate side-effect of having the same struct for both update and create. Unless I'm misunderstanding your comment.

application/gqlgen/schema.graphql
```diff
@@ -12,37 +12,25 @@ type Query {
     Posts, aka Requests. With no parameters supplied, all posts visible to the authenticated user are returned. Filter
     parameters only remove from this default list and never include posts that are not visible to the authenticated
     user. For posts associated with a `User` or `Meeting`, use the `posts` field on `User` and `Meeting`.
-    NOT YET IMPLEMENTED: `destination`, `origin`, `searchText`
     """
     posts(
         """
-        NOT YET IMPLEMENTED --
```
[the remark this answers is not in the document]
Schparky: Yep

application/models/post.go
Schparky: Good.

application/models/post.go
Schparky: Probably, yeah.

Schparky: Or what if I just empty it out instead of error it?

Baggerone: That sounds good.

application/models/post.go
Schparky: Good thought. It doesn't combine into one DB query, but a little simpler nonetheless.

application/models/user.go
```diff
@@ -746,3 +746,13 @@ func (u User) GetLanguagePreference() string {
 func (u *User) GetRealName() string {
 	return strings.TrimSpace(u.FirstName + " " + u.LastName)
 }
+
+// HasOrganization returns true if the user has one or more organization connections
+func (u *User) HasOrganization() bool {
+	var c Count
+	err := DB.RawQuery("SELECT COUNT(*) FROM user_organizations WHERE user_id = ?", u.ID).First(&c)
+	if err != nil || c.N == 0 {
```
Baggerone: Maybe log real errors, so we can troubleshoot if this doesn't work for some reason?

Schparky: OK

PR #244, opened 2020-02-14
add `posts` to `meeting` queries

Schparky:


application/migrations/20200214201745_drop_post_meeting_unique_index.up.fizz
```diff
@@ -0,0 +1 @@
+drop_index("posts", "posts_meeting_id_idx")
```
Baggerone: Oops. I must not have been thinking straight when I added this originally.

PR #246, opened 2020-02-14
include post title in notification message subject

Schparky:


application/locales/global.en.yaml
```diff
@@ -223,7 +223,7 @@
 
 # New Message notification subject
 - id: Email.Subject.Message.Created
-  translation: New message from {{.sentByNickname}} on {{.AppName}}
+  translation: {{.AppName}} message from {{.sentByNickname}} about {{.postTitle}}
```
Baggerone: I guess this broke some tests.

Schparky: I guess you can't have a leading `{` in a string.

PR #248, opened 2020-02-17
Add photoID tests

Baggerone:


application/models/post_test.go
```diff
@@ -1390,6 +1390,31 @@ func (ms *ModelSuite) TestPost_RemovePhoto() {
...
+
+	var photoFixture File
+	const filename = "photo.gif"
+	ms.Nil(photoFixture.Store(filename, []byte("GIF89a")), "failed to create file fixture")
+
+	attachedFile, err := post.AttachPhoto(photoFixture.UUID.String())
+	ms.NoError(err, "failed to attach photo to post")
+	ms.Equal(filename, attachedFile.Name)
+	ms.True(attachedFile.ID != 0)
+	ms.True(attachedFile.UUID.Version() != 0)
```
Schparky: Just as an FYI, not as a change suggestion, I saw this done in some Buffalo code:
```go
if uuid == uuid.Nil {
    // not a good UUID
}
```

PR #249, opened 2020-02-18
create meeting invitations

Schparky:


application/gqlgen/mutations.go
```diff
@@ -245,17 +248,52 @@ func (r *mutationResolver) RemoveOrganizationTrust(ctx context.Context, input Re
...
+		MeetingID: m.ID,
+		InviterID: cUser.ID,
+	}
+
+	badEmails := make([]string, 0)
+	for _, email := range input.Emails {
+		inv.Email = email
+		if err := inv.Create(); err != nil {
+			badEmails = append(badEmails, email)
+			domain.ErrLogger.Printf("error creating meeting invite for email '%s', %s", email, err)
```
Baggerone: You don't need to change this - I'm just wondering if it would be better to have one log message with all the bad emails together (down in the next `if` block).

application/models/meetinginvite.go
```diff
@@ -0,0 +1,69 @@
...
+
+// Create validates and stores the MeetingInvite data as a new record in the database.
+func (m *MeetingInvite) Create() error {
+	invite := *m
+	invite.Secret = domain.GetUUID()
+
+	err := create(&invite)
+	if err != nil && err.Error() ==
+		`pq: duplicate key value violates unique constraint "meeting_invites_meeting_id_email_idx"` {
+		err = nil
```
Baggerone: In this case, is it important to hydrate `invite` with all of the existing MeetingInvite's values?

application/gqlgen/schema.graphql
```diff
@@ -114,24 +114,21 @@ type Mutation {
     createOrganizationTrust(input: CreateOrganizationTrustInput!): Organization!
     removeOrganizationTrust(input: RemoveOrganizationTrustInput!): Organization!
 
-    """
-    NOT YET IMPLEMENTED --
-    Bulk create `MeetingInvitation`s
-    """
-    createMeetingInvitations(input: CreateMeetingInvitationsInput!): [MeetingInvitation!]!
+    "Bulk create `MeetingInvite`s and return the updated list of invites for the specified meeting."
```
Baggerone: Should we add something like ... `can be called multiple times with repeat email addresses being ignored.`?

application/models/meetinginvite.go
Schparky: Since it would involve a new database call, I think it's safe to wait to see if it's needed.

application/gqlgen/mutations.go
Schparky: I thought about that, but we would lose the actual failure message.

PR #250, opened 2020-02-18
fix neededBefore validation

Schparky:
This is a (possibly temporary) fix for the card "Unable to delete a post that is past it's neededBefore date". It moves the `neededBefore` validation to the `ValidateCreate` method so updates like setting `status` to `REMOVED` won't fail.

### DIFF
--- a/application/models/post.go
+++ b/application/models/post.go
@@ -352,14 +352,6 @@ func (p *Post) DestroyPotentialProviders(status PostStatus, user User) error {
 
 // Validate gets run every time you call a "pop.Validate*" (pop.ValidateAndSave, pop.ValidateAndCreate, pop.ValidateAndUpdate) method.
 func (p *Post) Validate(tx *pop.Connection) (*validate.Errors, error) {
-	tomorrow := time.Now().Truncate(domain.DurationDay).Add(domain.DurationDay)
-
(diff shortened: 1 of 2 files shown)


forevermatt:
Does this mean that we are no longer validating the neededBefore date when actually updating the Post (not merely changing its status)?

Schparky:
> Does this mean that we are no longer validating the neededBefore date when actually updating the Post (not merely changing its status)?

Yes, that's correct. I realize it's not the perfect solution, but a quick attempt to improve the situation.

forevermatt:
> Yes, that's correct. I realize it's not the perfect solution, but a quick attempt to improve the situation.

My gut reaction is that the original problem isn't big enough to merit a quick fix (rather than a long-term fix). However, as long as we make a note of this new bit of technical debt, I'm okay with this.

Schparky:
> > Yes, that's correct. I realize it's not the perfect solution, but a quick attempt to improve the situation.
> 
> My gut reaction is that the original problem isn't big enough to merit a quick fix (rather than a long-term fix). However, as long as we make a note of this new bit of technical debt, I'm okay with this.

OK, will make a note of it. On the other hand, I'm not sure this is really a problem. No harm can really come from setting the date back. Since the UI does validation, it should protect the user from an accidental mistake.

PR #252, opened 2020-02-21
Auth changes for meeting-based login

Baggerone:
Does not allow for social auth options yet.
At Billy's suggestion, I included the idea of an InviteType, in order to allow for future invites that are for things that aren't meetings.


application/models/post_test.go
```diff
@@ -1690,16 +1690,17 @@ func (ms *ModelSuite) TestPosts_FindByUser() {
 func (ms *ModelSuite) TestPosts_GetPotentialProviders() {
 	t := ms.T()
 
-	f := createFixturesFor_Posts_GetPotentialProviders(ms)
```
Schparky: can `createFixturesFor_Posts_GetPotentialProviders` be removed now?

application/actions/auth.go
```diff
@@ -30,6 +30,14 @@ const (
...
 	ExpiresUTCParam = "expires-utc"
 
+	// http params for the Invite type and code
+	InviteCode        = "code"
+	InviteTypeMeeting = "meeting"
+
+	// session keys for using invites for authentication
+	InviteTypeSessionKey       = "InviteType"
+	InviteObjectUUIDSessionKey = "InviteObjectID"
+
```
Schparky: Maybe prefix the http param names with something like `QueryParam` and the sessions keys with `SessionKey`. (The session keys have it at the end, which seems a bit different than our norm.)

Actually, I just looked at the surrounding code and the "norm" for this file is to have the "category" at the end, so maybe just add `Param` to the first two, unless you feel like changing all of them.

Baggerone: :+1:

application/gqlgen/schema.graphql
```diff
@@ -112,8 +107,8 @@ type Mutation {
     NOT YET IMPLEMENTED --
     Validate a `MeetingInvite` and create a new `MeetingParticipant`. If the `MeetingParticipant.User` contains
     one or more `Organizations`, the user should be redirected to the standard login page. Otherwise, login should
-    proceed through one of the configured `AuthType`s described by the `systemConfig` query.
```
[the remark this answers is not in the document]
Baggerone: OK.

application/models/post_test.go
Baggerone: :+1:

application/models/meetingparticipant.go
```diff
@@ -55,3 +55,14 @@ func (m *MeetingParticipant) Invite() (*MeetingInvite, error) {
 	}
 	return &invite, nil
 }
+
+func (m *MeetingParticipant) FindByMeetingIDAndUserID(meetingID, userID int) error {
+	return DB.Where("meeting_id = ? and user_id = ?", meetingID, userID).First(m)
+}
+
+func (m *MeetingParticipant) CreateForInvite(invite MeetingInvite, userID int) error {
```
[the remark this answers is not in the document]
Baggerone: OK

PR #255, opened 2020-02-25
provide a list of user's meetings

Schparky:
Also, I moved some helper functions so `models` code can use them -- specifically, so `models.User.MeetingsAsParticipant` can be used as a GraphQL resolver rather than adding a separate method in `gqlgen.userResolver`. Unfortunately I had to touch many files to do this, but I think it will pay off over time. Since `reportError` is being moved and renamed, is it still named well? Maybe `translateError`?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -94,7 +94,7 @@ return nil, domain.ReportError(ctx, err, "UpdatePostStatus.Unauthorized", extras
 Errors within a query, such as an authorization failure on a request field, should
 not present an error to the API client. The requested data field should be returned
 as `null`. The cause of the error should be logged to `stderr` and Rollbar using
-`domain.ReportError` if context is available, or `domain.ErrLogger.printf` if no
+`domain.ReportError` if context is available, or `domain.ErrLogger.Printf` if no
 context is available.
 
 #### REST API responses
@@ -120,7 +120,7 @@ err := domain.AppError{
 return c.Render(httpStatus, render.JSON(err))
 ```
 
-In addition, the error should be logged using `Error` or `ErrLogger.printf`
+In addition, the error should be logged using `Error` or `ErrLogger.Printf`
 
 #### Internal error logging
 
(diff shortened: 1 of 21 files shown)


CONTRIBUTING.md
```diff
@@ -70,24 +70,64 @@ situations. For example:
...
-the resolver function. Translation of error messages is handled by the Buffalo
-`Translate` function. Translation keys consist of an initial identifier related to 
-a model, query or mutation name, optionally followed by a short description of the 
-point of failure. You may use the helper function `reportError` which handles all of 
-these steps. Translation text is stored in the `locales` folder.
- 
+#### GraphQL error responses
+
+Errors occurring at the top level of a query or mutation, such as an authorization
+failure of an entire query, should present to the API client with an http OK (200)
```
Baggerone:
```suggestion
failure of an entire query, should present to the API client an http OK (200)
```

CONTRIBUTING.md
```diff
@@ -70,24 +70,64 @@ situations. For example:
...
+response with an `errors` object in the response body. Internally, this is handled
+by the helper function `domain.ReportError`, which takes an `error` and a translation
+key like `"UpdateUser.NotFound"`. The `error` is logged to `stderr` and to Rollbar,
+and the translation key is localized by Buffalo's `Translate` function and returned
+from the resolver to be processed by `gqlgen`. Translation text is stored in the
+`locales` folder.
```
Baggerone:
```suggestion
`application/locales` folder.
```

CONTRIBUTING.md
```diff
@@ -70,24 +70,64 @@ situations. For example:
...
+```go
+err := domain.AppError{
+    Code: httpStatus,
+    Key:  errorCode,
+}
+
+return c.Render(httpStatus, render.JSON(err))
+```
+
+In addition, the error should be logged using `Error` or `ErrLogger.Printf`
```
Baggerone: Do you want to add something about actions.authRequestError() doing this for you in that package?

Schparky: I chose not to since it's written just for `/auth` endpoints. It might be worth rewriting that helper so it's more general-purpose. AFAIK, there is no helper for the other REST endpoints.

PR #256, opened 2020-02-26
implement removeMeetingInvite mutation

Schparky:


application/actions/meeting_test.go
```diff
@@ -364,3 +371,73 @@ func (as *ActionSuite) Test_CreateMeetingInvites() {
...
+		{
+			Name:           "organizer",
+			Email:          "invitee1@example.com",
+			MeetingID:      f.Meetings[2].UUID.String(),
+			TestUser:       f.Users[0],
+			ResponseEmails: []string{},
+		},
+	}
+
+	for _, tc := range testCases {
```
Baggerone: I just noticed that this and some others aren't using the `t.Run(tc.name, func(t *testing.T) { ...})`  way of running the tests. 
I don't know if it's worth changing it.

application/locales/global.en.yaml
```diff
@@ -35,8 +35,18 @@
...
+- id: CreateMeetingInvite.Unauthorized
+  translation: You are not allowed to create invitations for that meeting
 - id: CreateMeetingInvite.ListInvites
   translation: We had a problem getting the list of invites after adding the new invites
+- id: RemoveMeetingInvite.FindMeeting
+  translation: We had a problem finding the meeting for the invite to be removed
+- id: RemoveMeetingInvite.Unauthorized
+  translation: You are not allowed to remove invitations for that meeting
+- id: RemoveMeetingInvite.ListInvites
+  translation: We had a problem getting the list of invites after removing the new invites
```
Baggerone:
```suggestion
  translation: We had a problem getting the list of invites after removing the invite
```

application/actions/meeting_test.go
Schparky: Good suggestion. I'll see what it would take to make that change.

application/locales/global.en.yaml
Schparky: haha, thanks.

PR #258, opened 2020-02-27
Multi auth login (org based and social based)

Baggerone:
This will need to be coordinated with the UI release to avoid breaking things.


application/models/user.go
```diff
@@ -146,6 +147,29 @@ func (u *User) CreateAccessToken(org Organization, clientID string) (string, int
 	return token, expireAt.UTC().Unix(), nil
 }
 
+// CreateOrglessAccessToken - Create and store new UserAccessToken with no associated UserOrg
+func (u *User) CreateOrglessAccessToken(clientID string) (string, int64, error) {
```
Schparky: What if you made the `org` parameter of `CreateAccessToken` a pointer or test for `ID == 0`? These two functions are nearly identical.

application/models/user.go
```diff
@@ -239,6 +263,55 @@ func (u *User) FindOrCreateFromAuthUser(orgID int, authUser *auth.User) error {
...
+	}
+
+	if newUser {
+		e := events.Event{
+			Kind:    domain.EventApiUserCreated,
+			Message: "Nickname: " + u.Nickname + "  UUID: " + u.UUID.String(),
+			Payload: events.Payload{"user": u},
+		}
+		emitEvent(e)
+	}
```
Schparky: Looks like 276-310 could be extracted, if not fully combining the two `FindOrCreate` methods.

application/models/user.go
```diff
@@ -239,6 +263,55 @@ func (u *User) FindOrCreateFromAuthUser(orgID int, authUser *auth.User) error {
...
 
+// FindOrCreateFromOrglessAuthUser creates a new User based on an auth.User and
+// sets its SocialAuthProvider field so they can login again in future.
+func (u *User) FindOrCreateFromOrglessAuthUser(authUser *auth.User, authType string) error {
+
+	if err := DB.Where("email = ?", authUser.Email).First(u); err != nil {
+		if domain.IsOtherThanNoRows(err) {
+			return errors.WithStack(err)
+		}
+	}
```
Schparky:
```suggestion
	if err := DB.Where("email = ?", authUser.Email).First(u); domain.IsOtherThanNoRows(err) {
		return errors.WithStack(err)
	}
```

application/models/user_test.go
```diff
@@ -821,6 +931,90 @@ func (ms *ModelSuite) TestUser_FindByID() {
...
+	}{
+		{
+			name:    "Good",
+			email:   f.Users[0].Email,
+			wantErr: "",
+		},
+		{
+			name:    "Bad",
+			email:   "bad@example.com",
+			wantErr: "sql: no rows in result set",
```
Schparky: I just learned about this built-in constant.
```suggestion
			wantErr: sql.ErrNoRows,
```

application/actions/authsocial.go
```diff
@@ -0,0 +1,355 @@
...
+}
+
+func addTwitterConfig(configs map[string]SocialAuthConfig) {
+	addConfig(AuthTypeTwitter, domain.Env.TwitterKey, domain.Env.TwitterSecret, configs)
+}
+
+// getSocialAuthConfigs returns a map of the enabled Social Auth Provider key-secret pairs
+// In theory, this could go into an init function, but there is already
+// one in render.go and it's a bit risky having multiple init functions.
+func getSocialAuthConfigs() map[string]SocialAuthConfig {
```
Schparky: Would it be possible to do this in an `init()`?

application/actions/authsocial.go
```diff
@@ -0,0 +1,355 @@
...
+	}
+
+	// set person on rollbar session
+	domain.RollbarSetPerson(c, authUser.ID, authUser.Nickname, authUser.Email)
+
+	return c.Redirect(302, getLoginSuccessRedirectURL(authUser, callbackValues.returnTo))
+}
+
+// Hydrates an AuthUser struct
+func getOrglessAuthUser(clientID string, user models.User) (AuthUser, error) {
```
Schparky: Looks identical to `getAuthUser` except for the initial call to get the access token. Can these be combined into one function?

application/actions/auth.go
```diff
@@ -262,9 +289,9 @@ func createAuthUser(
 		return AuthUser{}, err
 	}
 
-	var uos []AuthOrgOption
+	var uos []authOption
 	for _, uo := range user.Organizations {
-		uos = append(uos, AuthOrgOption{
+		uos = append(uos, authOption{
```
Schparky: Not sure what this does. I don't see any later access of `uos`. Does `authUser` need this list?

application/actions/authsocial.go
Baggerone: It sounds like you didn't read my comment. (-:

Or did you and you still think it should go in one?

application/actions/auth.go
```diff
@@ -146,60 +147,68 @@ func authInvite(c buffalo.Context) error {
...
-		if org.AuthType == "" {
-			return org, userOrgs, errors.New("unable to find organization by email domain")
-		}
+// no user_organization records yet, see if we have an organization for user's email domain
+func getOrgForNewUser(authEmail string) (models.Organization, error) {
+	var org models.Organization
+	if err := org.FindByDomain(domain.EmailDomain(authEmail)); err != nil {
+		return org, err
 	}
+	return org, nil
```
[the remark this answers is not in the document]
Baggerone: :+1:

application/actions/app.go
```diff
@@ -72,11 +72,12 @@ func App() *buffalo.App {
 		app.POST("/service", serviceHandler)
 
 		auth := app.Group("/auth")
-		auth.Middleware.Skip(setCurrentUser, authInvite, authRequest, authCallback, authDestroy, serviceHandler)
+		auth.Middleware.Skip(setCurrentUser, authInvite, authRequest, authSelect, authCallback, authDestroy, serviceHandler)
```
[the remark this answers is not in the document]
Baggerone: :+1:

application/actions/auth.go
```diff
@@ -218,18 +227,35 @@ func meetingAuthRequest(c buffalo.Context, authEmail string, extras map[string]i
 			extras)
 	}
 
-	// find org for auth config and processing
-	org, userOrgs, err := getOrgAndUserOrgs(authEmail, c)
+	// If the user already has organizational auth options, use them
```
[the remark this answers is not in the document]
Baggerone: There is also a difference in the call to `finishAuthRequestForSocialUser()` in the middle of `authRequest()`. I'm not sure if refactoring it would actually end up being preferable.

application/actions/authsocial.go
Schparky: TBH, I didn't read the comment -- sorry about that. I do think it would still be good to use `init` since the benefits outweigh the risks (which I think are none in this case).

application/actions/auth.go
```diff
@@ -252,7 +278,8 @@ func inviteAuthRequest(c buffalo.Context, authEmail, inviteType string) error {
 	return nil
 }
 
-func createAuthUser(
+// Hydrates an AuthUser struct
+func getAuthUser(
```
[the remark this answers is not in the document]
Baggerone: :+1:

application/actions/auth.go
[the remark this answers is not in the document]
Schparky: In my opinion, it would be preferable because of the 20-30 lines of duplicate code, but you're the one who spent all the time figuring this out. I'll leave it in your court.

application/actions/auth.go
```diff
@@ -146,60 +147,68 @@ func authInvite(c buffalo.Context) error {
...
 	}
+	return org, nil
+}
 
-	return org, userOrgs, nil
+type authError struct {
+	httpStatus int
+	errorCode  string
+	errorMsg   string
 }
```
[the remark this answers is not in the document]
Baggerone: It's a little complicated, because these are based on the `authRequestError` signature.
I think I'll leave `httpStatus` as is and change `errorCode` to `errorKey` here and in the `authRequestError` signature. 

Later, if we want, we can change the `authRequestError` signature to expect an `domain.AppError` param, instead of the `httpStatus` and `errorKey` params.

application/models/user_test.go
Baggerone: :+1: ... it also needs the `.Error()`  on the end to make it a string.

application/actions/auth.go
Baggerone: It must be left-over code from when we were trying to prepare for Users potentially having more than one org.

application/actions/authsocial.go
Baggerone: :+1:

application/actions/auth.go
Baggerone: I'll remove it.

application/models/user.go
Baggerone: :+1:

application/actions/authsocial.go
Baggerone: OK. How about a hybrid ...
I'll keep modified versions of the two functions for ease of testing and I'll only call them from the init() function.  I'll also give them each a doc comment, "// Forbidden. Do NOT call this function.  (Only called by init() and a test)"

Schparky: Sounds good. Is the "forbidden" warning necessary, since there's a check to see if it's already been called?

Baggerone: Well, I removed the check - I thought that was probably why you wanted to have them in the init() function - so we just execute them once.

Schparky: Oh, that makes sense.

application/actions/auth.go
```diff
@@ -291,33 +317,20 @@ func createAuthUser(
...
-	}
-
-	redirectURL, err := sp.AuthRequest(c)
-	if err != nil {
-		return authRequestError(c, http.StatusInternalServerError, domain.ErrorGettingAuthURL,
-			fmt.Sprintf("unable to determine what the authentication url should be for '%s' ... %v", org.Name, err))
+	for _, uo := range userOrgs {
+		option, authErr := getOrgBasedAuthOption(c, authEmail, uo.Organization)
+		if authErr != nil {
+			return authRequestError(c, authErr.httpStatus, authErr.errorCode, authErr.errorMsg, extras)
```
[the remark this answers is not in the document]
Baggerone: I'm not sure how much the map would help us.  But the first idea sounds OK.

  So, for example, this ...
```
	if err := meeting.FindByInviteCode(inviteCode); err != nil {
		return authInviteResponse{}, authRequestError(c, http.StatusNotFound, domain.ErrorInvalidInviteCode,
			"error validating Invite Code: "+err.Error(), extras)
	}
```
would become this ...
```
	if err := meeting.FindByInviteCode(inviteCode); err != nil {
		authErr := authError{
			httpStatus: http.StatusNotFound,
			errorKey:   domain.ErrorInvalidInviteCode,
			errorMsg:   "error validating Invite Code: " + err.Error(),
		}
		return authInviteResponse{}, authRequestError(c, authErr, extras)
	}
```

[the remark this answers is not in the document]
Schparky: Yes, that's what I was thinking. The map would do a couple things. Put the mapping between key and status code all in one place, and ensure consistency. For instance, I noticed that one of the keys (`domain.ErrorFindingOrgUserOrgs`) produced two different status codes. Not sure if that was intended or not. I guess the "all in one place" could also be a negative, so I could go either way on it.

application/models/user.go
Baggerone: :+1:

application/domain/domain.go
```diff
@@ -167,6 +168,7 @@ func init() {
 // readEnv loads environment data into `Env`
 func readEnv() {
 	Env.AccessTokenLifetimeSeconds = envToInt("ACCESS_TOKEN_LIFETIME_SECONDS", AccessTokenLifetimeSeconds)
+	Env.ApiBaseURL = envy.Get("API_BASE_URL", "")
 	Env.AppName = envy.Get("APP_NAME", "WeCarry")
 	Env.AuthCallbackURL = envy.Get("AUTH_CALLBACK_URL", "")
```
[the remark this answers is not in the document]
Baggerone: :+1:

application/models/user.go
Baggerone: :+1:

application/auth/google/provider.go
```diff
@@ -25,21 +25,25 @@ const ProviderName = "google"
...
+	// If jsonConfig is provided, use it. Otherwise, use the SocialAuthConfig
+	if len(jsonConfig) > 10 { // just some small number to see if it probably has valid data
+		if err := json.Unmarshal(jsonConfig, &config); err != nil {
+			return &Provider{}, err
+		}
+	}
 
-	if googleKey == "" || googleSecret == "" {
+	if config.Key == "" || config.Secret == "" {
 		err := errors.New("missing required environment variable for Google Auth Provider")
```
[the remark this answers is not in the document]
Baggerone: :+1:

application/auth/google/provider.go
```diff
@@ -25,21 +25,25 @@ const ProviderName = "google"
...
 // one manually.
-func New(jsonConfig json.RawMessage) (*Provider, error) {
+func New(config struct{ Key, Secret string }, jsonConfig json.RawMessage) (*Provider, error) {
 
-	googleKey := domain.Env.GoogleKey
-	googleSecret := domain.Env.GoogleSecret
+	// If jsonConfig is provided, use it. Otherwise, use the SocialAuthConfig
+	if len(jsonConfig) > 10 { // just some small number to see if it probably has valid data
+		if err := json.Unmarshal(jsonConfig, &config); err != nil {
+			return &Provider{}, err
```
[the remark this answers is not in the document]
Baggerone: :+1:

PR #259, opened 2020-02-27
refactor getters and setters

Schparky:
This is mostly just renames, with a couple noteworthy exceptions:
- In `Meeting` the image getter method handles hydration of the file object so calling code need not manage that. To accomplish that, the struct field `ImgFile` has been changed to a pointer to indicate whether it has been hydrated. (It could be non-existent as well, but I don't have a good answer for that at this point.) On a somewhat anecdotal note, "Effective Go" says that a field hidden by getter/setter methods should be lower-cased to avoid collision with the getter name. That doesn't work here because Pop needs access to the field.
- `Meeting.Posts()` is similar, but has no corresponding struct member. No setter is needed.
- same with `Organization.Domains()`
- same with `User.Posts()`

__Question:__ is the `Meeting.ImgFile` struct member necessary? Acts as a sort of "cache" but maybe has negligible benefit. I did this as a sort of trial, before doing the same thing elsewhere.

### DIFF
--- a/application/actions/auth.go
+++ b/application/actions/auth.go
@@ -126,7 +126,11 @@ func getAuthInviteResponse(c buffalo.Context) (authInviteResponse, error) {
 		Name: meeting.Name,
 	}
 	if meeting.ImageFileID.Valid {
-		resp.ImageURL = meeting.ImageFile.URL
+		f, err := meeting.ImageFile()
+		if err != nil {
+			domain.ErrLogger.Printf("error loading meeting image file: " + err.Error())
+		}
+		resp.ImageURL = f.URL
 	}
 	return resp, nil
 }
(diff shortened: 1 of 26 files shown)


Baggerone:
Unless there is a compelling reason not to do this, it seems important that we choose a different name for the getter relative to the field name. So `meeting.ImageFile` and `meeting.GetImageFile()` or something like that.

Schparky:
> Unless there is a compelling reason not to do this, it seems important that we choose a different name for the getter relative to the field name. So `meeting.ImageFile` and `meeting.GetImageFile()` or something like that.
> 
> I guess if we use a distinct name for the meeting getter, we should follow the same pattern for the other getters, even if they don't have a conflicting field name.

I was trying to go with the recommendation in [Effective Go](https://golang.org/doc/effective_go.html#Getters) by dropping the "Get". I don't see a strong case for keeping the field name, so that simplifies naming as well.

Baggerone:
I guess I'll wait for Phillip to chip in on this.

PR #262, opened 2020-03-03
createMeetingParticipant mutation

Schparky:
Implements `createMeetingParticipant` mutation.

Caveat: since the event visibility parameter has not been implemented, anyone can add themselves to any meeting.

Also in this PR:
- changed "meeting" to "event" in error message translation text
- `domain.Error()` doesn't panic if it is passed a context without an attached logger (e.g. a test context)
- `GetBuffaloContextFromGqlContext` renamed to `GetBuffaloContext` because it can take a `BuffaloContext` without panicking

### DIFF
--- a/application/models/meetingparticipant.go
+++ b/application/models/meetingparticipant.go
@@ -77,7 +77,7 @@ func (m *MeetingParticipant) CreateFromInvite(invite MeetingInvite, userID int) 
 // Otherwise, `code` must match either a MeetingInvite secret code or a Meeting invite code.
 func (m *MeetingParticipant) Create(ctx context.Context, meeting Meeting, code *string) error {
 	cUser := CurrentUser(ctx)
-	if !cUser.CanCreateMeetingParticipant(domain.GetBuffaloContextFromGqlContext(ctx), meeting, code) {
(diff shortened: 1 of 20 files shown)


application/models/meetingparticipant.go
```diff
@@ -72,3 +83,54 @@ func (m *MeetingParticipant) CreateFromInvite(invite MeetingInvite, userID int)
 func (m *MeetingParticipant) Destroy() error {
 	return DB.Destroy(m)
 }
+
+// Create a new MeetingParticipant from a meeting ID and code. If `code` is nil, the meeting must be non-INVITE_ONLY.
+// Otherwise, `code` must match either a MeetingInvite secret code or a Meeting invite code.
+func (m *MeetingParticipant) Create(ctx context.Context, meeting Meeting, code *string) error {
```
Baggerone: Should this be named `FindOrCreate` because of line 97?

Schparky: The goal was to avoid an error if the participant already exists. Maybe this means the method is trying to do too much. I think I'll pull out that "find" code and let the caller do it. Sound good?

Schparky: On second thought, I like your suggestion better. I can't think of any reason to have a `Create` method that fails when it tries to add a duplicate.

PR #271, opened 2020-03-06
make authType a GraphQL enum

Schparky:


application/models/organization.go
```diff
@@ -19,17 +22,51 @@ import (
...
 )
 
-const AuthTypeAzureAD = "azureadv2"
-const AuthTypeGoogle = "google"
-const AuthTypeSaml = "saml"
+type AuthType string
+
+const (
+	AuthTypeAzureAD AuthType = "AZUREADV2"
+	AuthTypeDefault AuthType = "DEFAULT"
```
Baggerone: Maybe add a comment about this one.  Does this mean that the organization's authType will be used rather than a special one for an OrgDomain?

Schparky: Yes, it replaces what is now just an empty string. I'll add a comment to clarify. By the way, I didn't extend this to the AuthTypes for user.social_auth_provider. Not sure if that should be defined in the same place or separately.

PR #275, opened 2020-03-10
Mark Request As Delivered/Received

Baggerone:


application/gqlgen/post.go
```diff
@@ -589,3 +589,76 @@ func (r *mutationResolver) RemovePotentialProvider(ctx context.Context, postID,
...
+		"user": cUser.UUID,
+		"post": post.UUID,
+	}
+
+	if cUser.AdminRole != models.UserAdminRoleSuperAdmin && cUser.ID != post.ProviderID.Int {
+		return nil, domain.ReportError(ctx, errors.New("not allowed to deliver request"),
+			"MarkRequestAsDelivered.Unauthorized", extras)
+	}
+
+	if post.Status != models.PostStatusAccepted && post.Status != models.PostStatusReceived {
```
Schparky: Shouldn't the post state machine check these? If not, perhaps it should. Seems like model validation rules leaking out. (Same on the other mutation.)

application/gqlgen/post.go
```diff
@@ -589,3 +589,76 @@ func (r *mutationResolver) RemovePotentialProvider(ctx context.Context, postID,
...
+	if err := post.FindByUUID(postID); err != nil {
+		return nil, domain.ReportError(ctx, err, "MarkRequestAsDelivered.FindPost")
+	}
+
+	extras := map[string]interface{}{
+		"user": cUser.UUID,
+		"post": post.UUID,
+	}
+
+	if cUser.AdminRole != models.UserAdminRoleSuperAdmin && cUser.ID != post.ProviderID.Int {
```
Schparky: Can this be done with `post.canUserChangeStatus` to keep all the logic on one place? (Same on the other mutation.)

application/gqlgen/schema.graphql
```diff
@@ -178,6 +177,12 @@ type Mutation {
     "Cancel a carry offer for any user. Authorized for the post creator, the potential provider, and Super Admins."
     removePotentialProvider(postID: String!, userID: String!): Post!
 
+    "Provider changes the status of a request to DELIVERED"
+    markRequestAsDelivered(postID: String!): Post!
+
+    "Requester changes the status of a request to RECEIVED"
+    markRequestAsReceived(postID: String!): Post!
+
```
Schparky: Are these effectively shortcuts or aliases to `updatePostStatus`? I'm not sure what this gains us.

application/gqlgen/post.go
```diff
@@ -589,3 +589,76 @@ func (r *mutationResolver) RemovePotentialProvider(ctx context.Context, postID,
...
+	cUser := models.CurrentUser(ctx)
+
+	var post models.Post
+	if err := post.FindByUUID(postID); err != nil {
+		return nil, domain.ReportError(ctx, err, "MarkRequestAsDelivered.FindPost")
+	}
+
+	extras := map[string]interface{}{
+		"user": cUser.UUID,
+		"post": post.UUID,
```
Schparky: These are somewhat redundant. The current user is added automatically by the rollbar code, and the post ID is in the mutation that is automatically logged. Feel free to keep them here if you like. (Same on other mutation.)

application/gqlgen/post.go
```diff
@@ -589,3 +589,76 @@ func (r *mutationResolver) RemovePotentialProvider(ctx context.Context, postID,
 
 	return &post, nil
 }
+
+func (r *mutationResolver) MarkRequestAsDelivered(ctx context.Context, postID string) (*models.Post, error) {
```
Schparky: Can this call the mutation resolver for UpdatePostStatus? Not only would it reduce duplicate code, but keep tests cleaner and shorter. Anything we can do to trim test execution time would be good.

application/gqlgen/schema.graphql
Baggerone: Matt was asking for them so that the UI doesn't need to know what goes on in the backend.  For example, for now a Post that is Accepted or Delivered goes straight to Completed if the creator says "I received it".  Some day, though, it may just go to Received until another step has taken place, e.g. the provider says that they got paid.

Schparky: Still seems like a status update with an internal rule that says some states can "auto-advance" to the next state. That would be reflected on the value in the `status` field in the response.

application/gqlgen/post.go
Baggerone: :+1:

application/gqlgen/post.go
Baggerone: Yep. I've refactored to avoid duplicate code. It does seem good, though, to keep the tests in there, in case some day we want to have those resolvers do different things.

application/gqlgen/post.go
Baggerone: :+1:

application/gqlgen/post.go
Baggerone: :+1:

PR #277, opened 2020-03-11
Separate org-based auth from social-auth in README

Schparky:


README.md
```diff
@@ -58,23 +64,8 @@ use `http:localhost` as the host for the WeCarry API, due to AzureAD's policies.
...
-```
-
-The two environment variables `FACEBOOK_KEY` and `FACEBOOK_SECRET` 
-will need to be set for the appropriate Facebook oauth account and application.
-
-For local development, if you are using `http`, then you will likely need to 
-use `http:localhost` as the host for the WeCarry API, due to Facebook's policies.
-(This affects the `AUTH_CALLBACK_URL` in the `.env` file and the `buffalo.environment.HOST` value
-in the docker-compose file.)
+<h4 id="google"> Google </h4>
```
Baggerone: Would it be better to leave this as markdown (####Google), since github will give it a link in the UI anyway.

README.md
```diff
@@ -33,15 +33,21 @@ may have earlier versions of these installed.
...
 will render json with a Code entry for the error. To see a list of possible codes, 
 refer to domain/errorcodes.go.  In particular, the related codes are those 
 that have a comment referring to actions.AuthRequest.
 
-### Office365/AzureAD
+### Organization Authentication
+
+Three types of authentication can be used for organization-based authentication:
+Azure AD, Google, and SAML. These are configured in an Organization record or 
+an Organizations Domain record, and are supported by the GraphQL API.
```
Baggerone:
```suggestion
an Organization Domain record, and are supported by the GraphQL API.
```

README.md
Schparky: There are two headings "Google" so rather that add extra word(s) I just did it this way. Do you know of another way to do it?

Schparky: Maybe I change it to Google GSuite?

Baggerone: Oh, ok. That should work.

PR #278, opened 2020-03-12
Randomize nickname prefixes and return better error messages about bad nicknames

Baggerone:


application/models/nicknames.go
```diff
@@ -34,3 +39,10 @@ func allPrefixes() [30]string {
 		"Confident",
 	}
 }
+
+func getShuffledPrefixes() [30]string {
+	ps := allPrefixes()
+	rand.Seed(time.Now().UnixNano())
+	rand.Shuffle(len(ps), func(i, j int) { ps[i], ps[j] = ps[j], ps[i] })
```
Schparky: cool!

Baggerone: Googled it :-)

PR #298, opened 2020-03-25
New Watch features

Schparky:
- added watch criteria to match the UI filters
- revised location matching to 100km radius and no country match
__NOTE__: post matching has not been updated -- only finds posts based on location. Will follow this PR up with another to fix this deficiency.


application/locales/global.en.yaml
```diff
@@ -307,14 +307,18 @@
 # Watch
 - id: GetWatchCreator
   translation: We had a problem finding the Watch creator
-- id: GetWatchLocation
-  translation: We had a problem finding the Watch location
+- id: GetWatchDestination
+  translation: We had a problem finding the Watch destination
```
fillup: These translations should all probably say `Alert` instead of Watch right?

application/gqlgen/schema.graphql
```diff
@@ -846,31 +846,65 @@ input UpdateUserPreferencesInput {
 }
 
 """
-A Watch for a given location. Requests with a destination near the watch location will trigger a notification to
-the watch creator. Other types of Watches (e.g. keyword search) may be created in future versions of WeCarry.
+A Watch for a given location. New requests matching all of the given criterium will generate a new
```
Baggerone: "... all of the given criteria ..."  (i.e. plural, right)?

Schparky: Yep.

application/gqlgen/watch.go
```diff
@@ -39,20 +39,46 @@ func (r *watchResolver) Owner(ctx context.Context, obj *models.Watch) (*PublicPr
...
+func (r *watchResolver) Destination(ctx context.Context, obj *models.Watch) (*models.Location, error) {
 	if obj == nil {
 		return &models.Location{}, nil
 	}
 
-	location, err := obj.GetLocation()
+	location, err := obj.GetDestination()
 	if err != nil {
-		return &models.Location{}, domain.ReportError(ctx, err, "GetWatchLocation")
+		return &models.Location{}, domain.ReportError(ctx, err, "GetWatchDestination")
```
Baggerone: This should return `nil, ...` because Destination is nullable, right?
(The same would apply to lines 45, 59 and 64)

Schparky: Good catch, thanks.

PR #300, opened 2020-03-27
notify on all watch criteria

Schparky:
Finishes the implementation of the new and improved Watch.

### DIFF
--- a/application/models/watch.go
+++ b/application/models/watch.go
@@ -150,15 +150,15 @@ func (w *Watch) Destroy() error {
 
 // matchesPost returns true if the Watch's Location is near the Post's Destination
 func (w *Watch) matchesPost(post Post) bool {
-	compareFunctions := []func(watch Watch, post Post) bool{
-		watchCompareSize,
-		watchCompareText,
-		watchCompareMeeting,
-		watchCompareDestination,
-		watchCompareOrigin,
+	compareFunctions := []func(*Watch, Post) bool{
+		(*Watch).compareSize,
+		(*Watch).compareText,
+		(*Watch).compareMeeting,
+		(*Watch).compareDestination,
+		(*Watch).compareOrigin,
 	}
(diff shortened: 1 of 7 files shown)


application/models/watch.go
```diff
@@ -147,30 +148,126 @@ func (w *Watch) Destroy() error {
...
+
+// matchesPost returns true if all non-null watch criteria match the post
 func (w *Watch) matchesPost(post Post) bool {
-	dest, err := post.GetDestination()
+	if w == nil {
+		domain.ErrLogger.Printf("nil receiver in Watch.matchesPost")
+		return false
+	}
+	compareFunctions := []func(*Watch, Post) bool{
+		(*Watch).compareSize,
```
Baggerone: Seems like these names would be more intuitive as `sizeMatches`, `textMatches` etc.
Normally, I think that "compare" refers to words like "smaller", "larger" etc. rather than "true"/"false".

application/models/watch.go
```diff
@@ -147,30 +148,126 @@ func (w *Watch) Destroy() error {
...
+}
+
+// matchesPost returns true if all non-null watch criteria match the post
 func (w *Watch) matchesPost(post Post) bool {
-	dest, err := post.GetDestination()
+	if w == nil {
+		domain.ErrLogger.Printf("nil receiver in Watch.matchesPost")
+		return false
+	}
+	compareFunctions := []func(*Watch, Post) bool{
```
Baggerone: Cool.  Where did you find out how to do this - with struct methods as a slice of functions?

Schparky: https://stackoverflow.com/questions/31561369/how-do-i-declare-a-function-pointer-to-a-method-in-go

application/models/watch.go
Schparky: Makes sense.

PR #306, opened 2020-04-01
Dataloaders NOT FOR MERGING

Baggerone:
Want to have a look over this and let me know if I'm going down the right path?


application/models/file.go
```diff
@@ -287,3 +287,9 @@ func (f *File) ClearLinked() error {
 	f.Linked = false
 	return DB.UpdateColumns(f, "linked")
 }
+
+// FindByIDs finds all Files associated with the given IDs and loads them from the database
+func (f *Files) FindByIDs(ids []int) error {
+	ids = domain.UniquifyIntSlice(ids)
```
Schparky, line 293: If this `UniquifyIntSlice` isn't done, PostgresQL will de-dupe them. Are you thinking this is better for performance?

application/actions/gql.go
```diff
@@ -19,6 +19,8 @@ func gqlHandler(c buffalo.Context) error {
 
 	h := handler.GraphQL(gqlgen.NewExecutableSchema(gqlgen.Config{Resolvers: &gqlgen.Resolver{}}))
 	newCtx := context.WithValue(c.Request().Context(), domain.BuffaloContext, c)
+	newCtx = dataloader.GetDataLoaderContext(newCtx)
```
Schparky, line 22: Does this double-nesting mess up functions like `CurrentUser()` and `GetBuffaloContext()`?

application/models/file.go
Baggerone: Are you saying Postgres will de-dupe the results or the integers going into the query?  If it's the results that get de-duplicated, wouldn't this function speed it up by reducing the numbers of integers in the where clause?

I don't really know myself.  It just looks like pop doesn't do any de-duping in advance.

application/actions/gql.go
Baggerone: It doesn't seem to -- all the tests pass.  I assume we have tests for those.  I can look into it a bit.

application/models/file.go
Schparky: I don't know about Pop or lib/pg but Postgres does de-dupe the results. I think you're right in that it would help with performance because the size of the query would be smaller, but maybe negligible. Just wanted to point it out in case you wanted to simplify. No problem either way.

application/actions/gql.go
Schparky: Good, no problem. I don't know if there's a test that would cover this, but the tests for those helper functions could be augmented to include a context with an extra layer.

application/gqlgen/meeting.go
```diff
@@ -92,11 +92,19 @@ func (r *meetingResolver) ImageFile(ctx context.Context, obj *models.Meeting) (*
...
+
+	image, err := dataloader.For(ctx).FilesByID.Load(obj.FileID.Int)
 	if err != nil {
 		return nil, domain.ReportError(ctx, err, "GetMeetingImage")
 	}
 
+	if err := image.RefreshURL(); err != nil {
+		return nil, err
+	}
+
```
Schparky, line 107: Did you weigh the alternative of doing this (RefreshURL) in the fetch method? At first glance, it seems that would be better. We might be able to do away with it altogether as it's only used in dev and test. So if we recognize and accept this limitation it could simplify production code.

application/models/file_test.go
```diff
@@ -203,6 +203,49 @@ func (ms *ModelSuite) TestFile_FindByUUID() {
 	}
 }
 
+func (ms *ModelSuite) TestFiles_FindByIDs() {
+	t := ms.T()
+
+	_ = createUserFixtures(ms.DB, 2)
```
Schparky, line 209: Is this call to `createUserFixtures` necessary or beneficial?

PR #313, opened 2021-05-10
run gofumpt on entire source tree

Schparky:


fillup (approved):

application/actions/authsocial_test.go
```diff
@@ -39,12 +39,18 @@ func (ts *TestSuite) Test_addConfig() {
...
 		want     map[string]SocialAuthConfig
 	}{
-		{name: "Facebook", authType: AuthTypeFacebook, key: fbKey, secret: fbSecret,
-			want: map[string]SocialAuthConfig{AuthTypeFacebook: fbConfig}},
-		{name: "Twitter", authType: AuthTypeTwitter, key: twKey, secret: twSecret,
-			want: map[string]SocialAuthConfig{AuthTypeTwitter: twConfig}},
-		{name: "Twitter", authType: AuthTypeLinkedIn, key: "", secret: twSecret,
-			want: map[string]SocialAuthConfig{}},
+		{
+			name: "Facebook", authType: AuthTypeFacebook, key: fbKey, secret: fbSecret,
```
jason-jackson, line 43: These probably would be more clean if split into more lines, but given their nature, I don't think it is necessary.

PR #318, opened 2021-05-11
remove file url_expiration

Schparky:
We decided a while back that URL expiration isn't necessary. 

Also did a bit of refactoring, which is why so many files were touched.


Baggerone (approved):
Wow. Lots of work.

Schparky:
> Wow. Lots of work.

Nah, just a bunch of copy-paste. If `CreateFileFixture` had existed prior to this, only a few files would have been touched.

Schparky:
> Wow. Lots of work.

Nah, just a bunch of copy-paste. If `CreateFileFixture` had existed prior to this, only a few files would have been touched.

Schparky:
After further thought, we probably want to keep the expiration

PR #328, opened 2021-06-10
changes to correspond with the UI switch to routify

Schparky:


application/templates/mail/request_from_open_to_accepted.plush.html
```diff
@@ -3,7 +3,7 @@ <h4><a href="<%= requestURL %>"><%= requestTitle %></a></h4>
 <p>
     <%= receiverNickname %> has accepted your offer to fulfill their request. Be sure to communicate with them
     to coordinate the details. For guidance on things to coordinate, see the
-    <a href="<%= uiURL %>/#/terms/responsibilities-of-providers">Responsibilities of Providers</a> section of our Terms of Use.
+    <a href="<%= uiURL %>/terms#responsibilities-of-providers">Responsibilities of Providers</a> section of our Terms of Use.
```
Baggerone, line 6: Is this right with a hash here between terms and responsibilities? (And next one)

Baggerone (approved):
Approved assuming you really wanted a hash after `/terms`

application/templates/mail/request_from_open_to_accepted.plush.html
Schparky: Yep. The slash was a workaround in the hash-based routing scheme. Couldn't have two hashes in the url.

PR #330, opened 2021-06-15
Release - use routify on UI

Schparky:
### Changed
- UI routes no longer need a hash (except for actual anchors) because of new UI router
- updated dependencies
- clean up unused location records


Baggerone (approved):
Approved, assuming this isn't causing the excessive db connections error :-)

hobbitronics (commented):
I can't say I understand much of it.

PR #331, opened 2021-06-16
add tx into call chain

Schparky:
- There are a few unrelated minor changes mixed in, I got on a brief tangent of cleaning up lint warnings
- I am not convinced the mix of `*pop.Connection` and `context.Context` parameters is a good thing. I've gone back and forth on this, and haven't come to a solid conclusion. Cannot use `context.Context` everywhere because some functions are called places that do not have a context. Cannot use `*pop.Connection` everywhere unless a handful of functions are changed to accept the current user. Also, context is helpful for logging extras and calling domain.Logger with context.

### DIFF
--- a/application/actions/authsocial.go
+++ b/application/actions/authsocial.go
@@ -132,7 +132,7 @@ func getSocialAuthOptions(authConfigs map[string]SocialAuthConfig) []authOption 
 // For non invite based ... based on a User in the database with a SocialAuthProvider value
 func finishAuthRequestForSocialUser(c buffalo.Context, authEmail string) error {
 	var user models.User
-	if err := user.FindByEmail(authEmail); err != nil {
+	if err := user.FindByEmail(models.DB, authEmail); err != nil {
 		authErr := authError{
 			httpStatus: http.StatusInternalServerError,
 			errorKey:   domain.ErrorFindingUserByEmail,
@@ -280,7 +280,7 @@ func socialLoginNonInviteBasedAuthCallback(c buffalo.Context, authEmail, authTyp
 	domain.NewExtra(c, "authType", authType)
 
 	var user models.User
-	if err := user.FindByEmailAndSocialAuthProvider(authEmail, authType); err != nil {
+	if err := user.FindByEmailAndSocialAuthProvider(models.DB, authEmail, authType); err != nil {
 		return logErrorAndRedirect(c, domain.ErrorGettingSocialAuthUser,
 			fmt.Sprintf("error loading social auth user for '%s' ... %v", authType, err))
 	}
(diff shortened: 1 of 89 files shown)


application/models/meeting.go
```diff
@@ -128,21 +128,21 @@ func (m *Meetings) FindOnDate(timeInFocus time.Time) error {
...
+	if err := getOrdered(m, Tx(ctx).Where("end_date >= ?", date)); err != nil {
 		return fmt.Errorf("error finding meeting with end_date before %s ... %s", date, err.Error())
 	}
 
 	return nil
 }
 
 // FindAfterDate finds the meetings that have a StartDate after the timeInFocus-date
-func (m *Meetings) FindAfterDate(timeInFocus time.Time) error {
+func (m *Meetings) FindAfterDate(tx *pop.Connection, timeInFocus time.Time) error {
```
Baggerone, line 142: Looks like this is only called from the tests. I see that it's different than FindOnOrAfterDate.

I wonder if they should have the same signature?

It looks like that is also true for FindOnDate.

Baggerone (approved):
Thanks for your work on this.

application/actions/authsocial.go
```diff
@@ -132,7 +132,7 @@ func getSocialAuthOptions(authConfigs map[string]SocialAuthConfig) []authOption
 // For non invite based ... based on a User in the database with a SocialAuthProvider value
 func finishAuthRequestForSocialUser(c buffalo.Context, authEmail string) error {
 	var user models.User
-	if err := user.FindByEmail(authEmail); err != nil {
+	if err := user.FindByEmail(models.DB, authEmail); err != nil {
```
fillup, line 135: Is this right? Same on line 283.

application/internal/test/test.go
```diff
@@ -175,7 +177,7 @@ func CreateFileFixture() models.File {
 		Name:    strconv.Itoa(rand.Int()) + ".gif",
 		Content: []byte("GIF89a"),
 	}
-	if err := f.Store(); err != nil {
+	if err := f.Store(models.DB); err != nil {
```
fillup, line 180: is this right? It's a test so maybe using global connection is okay?

application/job/job.go
```diff
@@ -42,11 +42,11 @@ func newThreadMessageHandler(args worker.Args) error {
 	}
 
 	var m models.Message
-	if err := m.FindByID(id, "SentBy", "Thread"); err != nil {
+	if err := m.FindByID(models.DB, id, "SentBy", "Thread"); err != nil {
```
fillup, line 45: Lots of `models.DB` in this file, is that right?

fillup: I'm thinking so if this runs as a background job and does not have a user context to pull a Tx from.

fillup (approved):

application/job/job.go
Schparky: Right

application/internal/test/test.go
Schparky: Probably, but I'll pass in the connection anyway.

application/actions/authsocial.go
Schparky: No, good find.

application/models/meeting.go
Schparky: Changed, thanks.

PR #332, opened 2021-06-17
make models signatures consistent

Schparky:
found it to be confusing having some function accept context.Context and some accept *pop.Connection, so switched all of them over to *pop.Connection


Baggerone (approved):
I'm curious why you added the MeetingsAsParticipant gql resolver.  Did we not need it before but now we do? Was the api broken before in that respect?

fillup (approved):

Schparky:
> I'm curious why you added the MeetingsAsParticipant gql resolver. Did we not need it before but now we do? Was the api broken before in that respect?

gqlgen can use functions in the `models` package as resolvers if the signature is correct. That's how we had done a couple of them. However, since I wanted to make every `models` function use tx instead of ctx, I had to move the resolver to the `gqlgen` package.

PR #333, opened 2021-06-18
Pulled over TTError stuff and added a rest endpoint for conversations

Baggerone:


Baggerone:
Need to have api.Error code first

PR #334, opened 2021-06-18
move ApiError into a new package and expand its capabilities

Schparky:
This is the error implementation I think we can use. The struct composition for models is still TBD. I nearly got it to work, then hit runtime errors from Pop. Will work on that separately.


Baggerone (approved):

CONTRIBUTING.md
```diff
@@ -112,7 +112,7 @@ or 500-level http response with a json body containing a `code` and a `key`:
 This can be generated as follows:
 
 ```go
-err := domain.AppError{
+err := apitypes.AppError{
```
fillup, line 115:
```suggestion
err := api.AppError{
```

fillup (approved):

PR #335, opened 2021-06-21
rest endpoint for conversations

Baggerone:
Modified to use new api package.


application/actions/app.go
```diff
@@ -66,6 +66,8 @@ func App() *buffalo.App {
 
 		app.POST("/gql/", gqlHandler)
 
+		app.GET("/conversations/", usersThreads)
```
fillup, line 69: do we need trailing slash here? Does that then require it on the API?

application/actions/messages.go
```diff
@@ -0,0 +1,28 @@
...
+
+func convertMessagesToAPIType(messages models.Messages) (api.Messages, error) {
+	var output api.Messages
+	if err := api.ConvertToOtherType(messages, &output); err != nil {
+		err = errors.New("error converting messages to apitype.messages: " + err.Error())
+		return nil, err
+	}
+
+	// Hydrate the thread's messages
+	for i := range messages {
```
fillup, line 18: I think this would be a little more clear to me if we iterated through `output` and then updated the current index `SentBy`. It may be safe to assume same ordering between original `messages` and `output`, but it'd be easier to trust and read if it was updating the thing it was iterating through.

application/actions/threads.go
```diff
@@ -0,0 +1,63 @@
...
+	"net/http"
+
+	"github.com/gobuffalo/buffalo"
+	"github.com/gobuffalo/buffalo/render"
+
+	"github.com/silinternational/wecarry-api/api"
+	"github.com/silinternational/wecarry-api/models"
+)
+
+// uploadHandler responds to POST requests at /upload
```
fillup, line 14:
```suggestion
// userThreads responds to GET requests at /conversations
```

application/api/messages.go
```diff
@@ -0,0 +1,37 @@
...
+import (
+	"time"
+
+	"github.com/gofrs/uuid"
+)
+
+// swagger:model
+type Messages []Message
+
+// A Message is one element associated with a conversation thread
```
Schparky, line 12: You could pull the docs from the graphql schema. I think they're a little more descriptive.

Schparky (commented):
Just a few comments, but I think the bigger question is how to handle the response structs.

Schparky (approved):

application/actions/app.go
Baggerone: :+1:  (I just did like the others)

application/actions/messages.go
Baggerone: :+1:

application/actions/threads.go
Baggerone: :+1:

application/models/thread.go
```diff
@@ -70,45 +71,65 @@ func (t *Thread) FindByUUID(tx *pop.Connection, uuid string) error {
...
+		}
 	}
 
-	return messages, nil
+	t.Messages = messages
+	return nil
 }
 
-func (t *Thread) GetParticipants(tx *pop.Connection) ([]User, error) {
+func (t *Thread) LoadParticipants(tx *pop.Connection) error {
```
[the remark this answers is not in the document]
Baggerone: :+1:

application/models/thread.go
```diff
@@ -70,45 +71,65 @@ func (t *Thread) FindByUUID(tx *pop.Connection, uuid string) error {
...
 }
 
-func (t *Thread) Messages(tx *pop.Connection) ([]Message, error) {
+func (t *Thread) LoadMessages(tx *pop.Connection, eagerFields ...string) error {
 	var messages []Message
-	if err := tx.Where("thread_id = ?", t.ID).All(&messages); err != nil {
-		return messages, fmt.Errorf("error getting messages for thread id %v ... %v", t.ID, err)
+	// If no eagerFields, then don't use Eager at all, otherwise it uses Eager on all of them
+
+	if len(eagerFields) > 0 {
```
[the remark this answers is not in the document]
Baggerone: :+1:

application/actions/threads.go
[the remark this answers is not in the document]
Baggerone: :+1:

application/api/messages.go
Baggerone: You mean this?

`Return a specific message. If the message is not visible to the auth user, an error will be returned.`

PR #336, opened 2021-06-21
/users/me endpoint

Schparky:


Baggerone (approved):
Approved with the understanding that you'll work on tests for it before too long. :-)

Schparky:
As @fillup pointed out, this approach doesn't help with input data.

PR #338, opened 2021-06-23
Adding initial /requests API endpoint and request resolving code. Nes…

chamberlainpj:
…ted objects currently returning as null, and less fields are returned in the Requests API data type than we should be returning, but it's a start.


Schparky (commented):
Definitely on the right track! You're right about adding fields to `api.Request` to fill out the things that the UI is expecting to receive.

As for how to do that, I usually like to see little functions broken out of bigger ones where possible and practical to do so. However, sometimes it seems these `actions` functions tend to grow because there's often no easy way to break pieces out. The `convertToOtherType` function will take care of most fields, but as you've seen, some things don't come through the other side.

application/actions/app.go
```diff
@@ -96,6 +96,8 @@ func App() *buffalo.App {
 
 		app.GET("/conversations", usersThreads)
 
+		app.GET("/requests", getRequests)
```
fillup, line 99: I think it'd be better to rename the handler method to `requestsList` so the resource comes first and the action second

Schparky: I might agree with that, but it's not how we did it in TF. So much standardization work to do, uff-da!

application/actions/requests.go
```diff
@@ -4,25 +4,92 @@ import (
...
+		output[i], err = convertRequestToAPIType(c, request)
+		if err != nil {
+			return []api.Request{}, err
+		}
+	}
+
+	return output, nil
+}
+
+func convertRequestToAPIType(c context.Context, request models.Request) (api.Request, error) {
```
Schparky, line 63:
```suggestion
func convertRequestToAPIType(ctx context.Context, request models.Request) (api.Request, error) {
```

application/actions/requests.go
```diff
@@ -4,25 +4,92 @@ import (
...
 	var output api.Request
 	if err := api.ConvertToOtherType(request, &output); err != nil {
 		err = errors.New("error converting request to api.request: " + err.Error())
 		return api.Request{}, err
 	}
 
 	// Hydrate the request's CreatedBy user
-	outputUser, err := convertUserToAPIType(ctx, request.CreatedBy)
+	outputUser, err := convertUserToAPIType(c, request.CreatedBy)
```
Schparky, line 71:
```suggestion
	outputUser, err := convertUserToAPIType(ctx, request.CreatedBy)
```

application/actions/requests.go
```diff
@@ -4,25 +4,92 @@ import (
...
+	tx := models.Tx(c)
+	destination, err := request.GetDestination(tx)
+	if err != nil {
+		err = errors.New("error converting request created_by user: " + err.Error())
+		return api.Request{}, err
+	}
+
+	var dest api.Location
+	if err := api.ConvertToOtherType(destination, &dest); err != nil {
+		err = errors.New("error converting request to api.request: " + err.Error())
```
Schparky, line 89:
```suggestion
		err = errors.New("error converting destination to api.Location: " + err.Error())
```

Schparky (commented):
Looks good. The only thing I might suggest is that, if there are several things to hydrate or load, make a function in `models` to do that. I could probably be convinced such a function should be in `actions` though, as I haven't given it a lot of thought.

application/actions/requests.go
```diff
@@ -4,25 +4,168 @@ import (
...
+	output,  err := convertRequestsToAPIType(c, requests)
+	if err != nil {
+		return reportError(c, appErrorFromErr(err))
+	}
+
+	return c.Render(200, render.JSON(output))
+}
+
+// converts list of model.Request into api.RequestAbridged
+func convertRequestsToAPIType(ctx context.Context, requests []models.Request) ([]api.RequestAbridged, error) {
```
Schparky, line 47:
```suggestion
func convertRequestsAbridged(ctx context.Context, requests []models.Request) ([]api.RequestAbridged, error) {
```

Schparky (commented):
Looks good, just have a few minor changes to recommend.

Schparky (approved):
I'll go ahead and mark it approved since my suggestions are minor and I'll be out the rest of the day.

Baggerone (approved):

application/actions/requests.go
```diff
@@ -24,3 +28,49 @@ func convertRequestToAPIType(request models.Request) (api.Request, error) {
...
+			return []api.Request{}, err
+		}
+	}
+
+	return output, nil
+}
+
+// getRequests responds to Get requests at /requests
+
+// go back to GraphQL code
```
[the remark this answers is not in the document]
chamberlainpj: That was a comment that I've since removed (didn't mean to commit it) more a reminder to me to look at the request models code/graphql resolvers for guidance on fetching requests from the database.

application/actions/requests.go
```diff
@@ -24,3 +28,49 @@ func convertRequestToAPIType(request models.Request) (api.Request, error) {
...
+
+	output,  err := convertRequestsToAPIType(requests)
+
+	if err != nil {
+		return reportError(c, appErrorFromErr(err))
+	}
+
+	return c.Render(200, render.JSON(output))
+	
+	// Returns all request information, but still lacking depth ("hydration") of CreatedBy,
```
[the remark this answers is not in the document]
chamberlainpj: Good to know! (Again, this was a comment removed in a subsequent commit after I did add this functionality.)

application/actions/requests_test.go
```diff
@@ -315,6 +317,39 @@ func (as *ActionSuite) Test_RequestsQuery() {
...
+	body := res.Body.String()
+	as.Equal(200, res.Code, "incorrect status code returned, body: %s", body)
+
+	var requestsList []api.Request
+	json.Unmarshal([]byte(body), &requestsList)
+	as.Equal(len(requests), len(requestsList))
+
+	wantContains := []string{}
+	for i := range requests {
+		wantContains = append(wantContains, fmt.Sprintf(`"id":"%s"`, requests[i].UUID))
```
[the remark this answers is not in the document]
chamberlainpj: Thanks for the syntactic sugar!

application/api/threads.go
```diff
@@ -35,7 +35,7 @@ type Thread struct {
 	Participants Users `json:"participants"`
 
 	// Request that owns this message thread
-	Request *Request `json:"request"`
+	Request *RequestAbridged `json:"request"`
```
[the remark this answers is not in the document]
chamberlainpj: Using api.Request in threads and api.RequestAbridged in requests does require slight code duplication or a significant increase in code complexity/logic to make things generic and eliminate the need for duplication. For now, I've taken the simpler approach of duplicating code that loads fields common to both request types, but we may want to revisit this in the future.

application/actions/requests_fixtures_test.go
```diff
@@ -83,6 +88,25 @@ func createFixturesForRequestQuery(as *ActionSuite) RequestQueryFixtures {
...
+	requests := test.CreateRequestFixtures(as.DB, 5, false)
+
+	requests[0].Status = models.RequestStatusAccepted
+	requests[0].ProviderID = nulls.NewInt(usersFixtures.Users[1].ID)
+	as.NoError(as.DB.Save(&requests[0]))
+
+	requests[1].Status = models.RequestStatusCompleted
+	requests[1].CompletedOn = nulls.NewTime(time.Now())
+	requests[1].ProviderID = nulls.NewInt(usersFixtures.Users[2].ID)
+	as.NoError(as.DB.Save(&requests[2]))
```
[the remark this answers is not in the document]
chamberlainpj: Nice to know :-)

PR #339, opened 2021-06-24
Threads - MarkAsRead

Baggerone:
I also changed the names of some of the actions test files and fixtures to be in the plural.  
There are still some test files that have a plural and singular form.  I didn't mess with those.


application/actions/threads.go
```diff
@@ -11,7 +12,18 @@ import (
...
+//
+// List the User's Conversations
+//
+//
+// ---
+// responses:
+//   '200':
+//     description: A list of the user's conversations with their messages
+//     schema:
+//       items:
```
Schparky, line 25: I think you need a `//       type: array` here.

application/actions/threads_test.go
```diff
@@ -0,0 +1,133 @@
...
+
+	req := as.JSON("/markMessagesAsRead")
+	req.Headers["Authorization"] = fmt.Sprintf("Bearer %s", users0.Nickname)
+	req.Headers["content-type"] = "application/json"
+	res := req.Post(reqBody)
+
+	body := res.Body.String()
+	as.Equal(200, res.Code, "incorrect status code returned, body: %s", body)
+
+	//wantContains := fmt.Sprintf(`"last_viewed_at":"%s`, testTime.Format("2006-01-02T15:04:05"))
```
Schparky, line 120: delete this comment?

application/actions/threads_test.go
```diff
@@ -0,0 +1,133 @@
...
+
+	users0 := f.Users[0]
+	testTime := time.Now().Add(1)
+
+	reqBody := api.MarkMessagesAsReadInput{
+		ThreadID: f.Threads[0].UUID.String(),
+		Time:     testTime,
+	}
+
+	req := as.JSON("/markMessagesAsRead")
```
Schparky, line 112: change this if the endpoint route changes

application/api/threads.go
```diff
@@ -16,20 +20,39 @@ type Thread struct {
 	// swagger:strfmt uuid4
 	// unique: true
 	// example: 63d5b060-1460-4348-bdf0-ad03c105a8d5
-	ID uuid.UUID `json:"uuid"`
+	ID uuid.UUID `json:"id"`
+
+	// CreatedAt - the time this thread was started
+	CreatedAt time.Time `json:"created_at"`
```
Schparky, line 26: From what I see, `CreatedAt` isn't used by the UI. Is there another reason to include it here?

application/api/users.go
```diff
@@ -5,12 +5,9 @@ import (
 	"github.com/gofrs/uuid"
 )
 
-// swagger:model
-type Users []User
-
 // Attributes of a user of the App
```
Schparky, line 8: maybe add to the comment to differentiate from the "public" type. Maybe "Full list of user attributes, including private information." or something like that.

application/models/thread.go
```diff
@@ -116,6 +119,16 @@ func (t *Thread) LoadParticipants(tx *pop.Connection) error {
 	if err := tx.Load(t, "Participants"); err != nil {
 		return fmt.Errorf("error loading threads participants %v ... %v", t.ID, err)
 	}
+
+	return nil
+}
+
+func (t *Thread) LoadThreadParticipants(tx *pop.Connection) error {
+
+	if err := tx.Load(t, "ThreadParticipants"); err != nil {
```
Schparky, line 128: Does this work? There isn't a `ThreadParticipants` field on `Thread`.

application/actions/app.go
```diff
@@ -95,6 +95,7 @@ func App() *buffalo.App {
 		app.POST("/gql/", gqlHandler)
 
 		app.GET("/conversations", usersThreads)
+		app.POST("/markMessagesAsRead", threadsMarkMessagesAsRead)
```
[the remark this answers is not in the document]
Baggerone: Maybe. Do we need to be concerned with sticking to the current gql query spelling or can we ignore that?

application/actions/threads.go
Baggerone: It worked without it -- maybe since Threads is a slice.

application/api/threads.go
Baggerone: :+1:

application/actions/app.go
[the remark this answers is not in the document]
fillup: I agree, I think it'd be better to see all thread related stuff under a common path. I'd expect endpoints something like these, though there may be better names/structures. 

```
// list all my threads
GET /threads

// get specific thread
GET /threads/{id}

// mark thread as read
PUT /threads/{id}/read

// add message to thread
POST /threads/{id}
```

application/actions/requests.go
```diff
@@ -1,26 +1,28 @@
...
 import (
+	"context"
 	"errors"
 
 	"github.com/silinternational/wecarry-api/api"
 	"github.com/silinternational/wecarry-api/models"
 )
 
-func convertRequestToAPIType(request models.Request) (api.Request, error) {
+func convertRequestToAPIType(c context.Context, request models.Request) (api.Request, error) {
```
chamberlainpj, line 11: nit: it seems like we often use ctx for context, but sometimes c. Do we care about consistency in naming, or does that not really matter?

application/actions/threads.go
```diff
@@ -25,15 +37,86 @@ func usersThreads(c buffalo.Context) error {
...
+	output, err := convertThreadsToAPIType(c, threads)
 	if err != nil {
 		return reportError(c, appErrorFromErr(err))
 	}
 
 	return c.Render(200, render.JSON(output))
 }
 
-func convertThreadsToAPIType(threads models.Threads) (api.Threads, error) {
+// swagger:operation POST /projects Threads MarkMessagesAsRead
```
fillup, line 48: This comment doesn't look right

application/models/thread.go
Baggerone: I think I did it this way because of your previous comment about making use of the many-to-many relationship.

It doesn't look like it's being called now.  I'll check if we need it.

application/actions/threads.go
Baggerone: :+1:

application/api/users.go
chamberlainpj: I think the User is now the "public" type, whereas "UserPrivate" is "private", but the comments should indeed be more clear as to which role the User fulfills

application/actions/threads_test.go
Baggerone: :+1:

application/api/threads.go
fillup: The UI doesn't currently use it, but it might at some point right? If it isn't an expensive field to include and there isn't a security/sensitivity issue with it is there any reason not to include it?

application/actions/threads_test.go
```diff
@@ -0,0 +1,133 @@
+package actions
```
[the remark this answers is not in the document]
Baggerone: :+1:

application/actions/app.go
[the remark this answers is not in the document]
Schparky: > Do we need to be concerned with sticking to the current gql query spelling or can we ignore that?

That can be ignored. That much will have to change in the UI code anyway, since GQL query name doesn't translate to the REST call.

application/api/threads.go
Schparky: Just trying to streamline where possible, but no, there's no reason I can think of to remove it.

application/actions/requests.go
Schparky: Yes, I think it matters, especially when we can easily get mixed up between `buffalo.Context` and `context.Context`. I believe the norm is `ctx` for `context.Context` and `c` for `buffalo.Context` (FWIW, I would like to see `bc` for `buffalo.Context` but that should probably not be changed at this point in time.)

Schparky (approved):

PR #340, opened 2021-06-25
my watches rest endpoint

Baggerone:


application/actions/organizations.go
```diff
@@ -9,7 +9,7 @@ import (
 
 func convertOrganizationsToAPIType(organizations models.Organizations) ([]api.Organization, error) {
 	output := make([]api.Organization, len(organizations))
-	for i := range organizations {
+	for i := range output {
```
Schparky, line 12: I am not sure what difference this makes. I think it was a suggestion by @fillup , right?

application/actions/watches.go
```diff
@@ -0,0 +1,91 @@
...
+		return reportError(c, appErrorFromErr(err))
+	}
+
+	return c.Render(200, render.JSON(output))
+}
+
+func convertWatchesToAPIType(ctx context.Context, tx *pop.Connection, watches models.Watches, user models.User) (api.Watches, error) {
+	var output api.Watches
+
+	if err := api.ConvertToOtherType(watches, &output); err != nil {
```
Schparky, line 51: Would it be too inefficient to convert one watch at a time in a function called `convertWatchToAPIType`? That way, if only one Watch is needed in a later endpoint, you would have consistency by calling the same function. This was done in organizations.go

application/actions/watches.go
```diff
@@ -0,0 +1,91 @@
...
+		err = errors.New("error converting watches to api.Watches: " + err.Error())
+		return nil, err
+	}
+
+	// Hydrate the watches' own and related fields
+	for i := range output {
+		if err := watches[i].LoadForAPI(tx, user); err != nil {
+			err = errors.New("error converting watch to api.Watch: " + err.Error())
```
Schparky, line 59:
```suggestion
			err = errors.New("error hydrating watch: " + err.Error())
```

application/actions/watches_test.go
```diff
@@ -0,0 +1,95 @@
...
+	"time"
+
+	"github.com/gobuffalo/nulls"
+
+	"github.com/silinternational/wecarry-api/domain"
+	"github.com/silinternational/wecarry-api/internal/test"
+	"github.com/silinternational/wecarry-api/models"
+)
+
+type watchQueryFixtures struct {
```
Schparky, line 14:
```suggestion
type watchFixtures struct {
```
and keep the original `watchQueryFixtures` untouched. "Query" isn't a word I would normally associate with a REST API.

application/api/meetings.go
```diff
@@ -0,0 +1,8 @@
+package api
+
+// MeetingName just gives the name of a meeting a/k/a Event, to serve as a focal point for finding, answering, carrying, and exchanging requests
+// swagger:model
+type MeetingName struct {
```
Schparky, line 5: I think we could use a full `Meeting` struct even though the UI only uses the name of the meeting when it's a child object of `Watch`. The `LoadMeeting` function loads all of the non-struct fields anyway. Might as well use them. Otherwise, if performance is a concern, we could get just the meeting name from the database.

application/models/watch.go
```diff
@@ -147,15 +159,31 @@ func (w *Watch) Destroy(tx *pop.Connection) error {
...
 	meeting := &Meeting{}
 	if err := tx.Find(meeting, w.MeetingID); err != nil {
-		return nil, err
+		return err
+	}
+	w.Meeting = meeting
+	return nil
+}
+
+// LoadForAPI assumes w.Destination and w.Origin were alread hydrated by an Eager query
```
Schparky, line 175: Why couldn't the Meeting be included in the Eager query?

application/actions/organizations.go
Baggerone: Phillip said it was easier for him to understand this way, so I thought I would try to make it consistent.

application/actions/watches.go
Baggerone: :+1:

application/actions/watches_test.go
Baggerone: :+1:

application/api/meetings.go
Baggerone: I would rather not include the meeting's related fields.  Is that OK?

application/models/watch.go
Baggerone: :+1:

application/actions/watches.go
Baggerone: :+1:

fillup (approved):
I didn't see anything beyond Schparky's comments

application/api/meetings.go
Schparky: Do you mean not to include struct fields on Meeting (e.g. File, Location) or not to include any fields except the name?

Schparky (approved):
looks good

PR #344, opened 2021-06-29
Create message Rest endpoint

Baggerone:


application/models/message_test.go
```diff
@@ -195,6 +196,82 @@ func (ms *ModelSuite) TestMessage_Create() {
...
+			threadUUID:  nil,
+			content:     "another bad message",
+			wantErr:     true,
+		},
+		{
+			name:        "bad, thread not valid for request",
+			user:        f.Users[0],
+			requestUUID: f.Requests[2].UUID.String(),
+			threadUUID:  &threadUUID,
+			content:     "another bad message",
```
chamberlainpj, line 250: Making the content distinct, e.g. "yet another bad message" might be helpful for quicker debugging/isolating which case fails, though the distinct names are almost certainly sufficient so it's probably not worth changing.

application/models/message.go
```diff
@@ -183,6 +185,90 @@ func (m *Message) Create(tx *pop.Connection, user User, requestUUID string, thre
...
+		appError := api.AppError{
+			Category: api.CategoryForbidden,
+			Key:      api.MessageRequestNotVisible,
+			Err:      errors.New("user cannot create a message on request"),
+		}
+		return &appError
+	}
+
+	var thread Thread
+	if input.ThreadID != nil && *input.ThreadID != "" {
```
chamberlainpj, line 218: Out of curiosity, why it is necessary to check for both non-nil and non-empty string here?

application/actions/threads.go
```diff
@@ -129,26 +129,43 @@ func convertThreadsToAPIType(ctx context.Context, threads models.Threads) (api.T
 
 	// Hydrate the thread's messages, participants
 	for i := range output {
-		messagesOutput, err := convertMessagesToAPIType(ctx, threads[i].Messages)
+		newThread, err := convertThread(ctx, threads[i])
```
chamberlainpj, line 132: Good to break this out into its own function :)

chamberlainpj (commented):
The message endpoint logic seems solid to me; I'm still pretty new to the repo though, so a second opinion would be good.

application/models/message.go
Baggerone: I guess there's a chance a non-nil but blank ThreadID could come through, in which case we want to ignore it.

If you try to just use the second condition on a nil pointer (without the first condition), it will panic.

application/models/message.go
```diff
@@ -183,6 +185,90 @@ func (m *Message) Create(tx *pop.Connection, user User, requestUUID string, thre
 	return nil
 }
 
+// Todo Once gql is no longer supported, simplify the name of this function and get rid of the Create() function above
+// Create a new message if authorized.
+func (m *Message) CreateForRest(tx *pop.Connection, user User, input api.MessageInput) *api.AppError {
```
fillup, line 190: I see your todo, so perhaps this is moot, but another option would be to name this method `CreateFromInput` which could stick around if there was need for another kind of create method.

fillup (approved):

application/models/message_test.go
Baggerone: :+1:

application/models/message.go
```diff
@@ -183,6 +185,90 @@ func (m *Message) Create(tx *pop.Connection, user User, requestUUID string, thre
...
+	} else if !isVisible {
+		appError := api.AppError{
+			Category: api.CategoryForbidden,
+			Key:      api.MessageRequestNotVisible,
+			Err:      errors.New("user cannot create a message on request"),
+		}
+		return &appError
+	}
+
+	var thread Thread
```
Schparky, line 217: Optional suggestion: wondering if this `if` block can/should be a separate function, if only to reduce the size of a long-ish function.

Schparky (approved):

PR #346, opened 2021-07-01
Feature/create watch

Baggerone:


application/actions/watches.go
```diff
@@ -6,12 +6,96 @@ import (
...
+//     required: true
+//     schema:
+//       "$ref": "#/definitions/WatchInput"
+// responses:
+//   '200':
+//     description: {"id": "<the id of the new watch>"}
+func watchesCreate(c buffalo.Context) error {
+	var input api.WatchInput
+	if err := StrictBind(c, &input); err != nil {
+		return reportError(c, &api.AppError{
```
Schparky, line 35: Do you want to merge in the latest from `dev2` so you can use `NewAppError`?

application/actions/watches.go
```diff
@@ -113,3 +197,47 @@ func convertWatch(tx *pop.Connection, watch models.Watch, user models.User) (api
...
+		if err := meeting.FindByUUID(tx, *input.MeetingID); err != nil {
+			return watch, err
+		}
+		watch.MeetingID = nulls.NewInt(meeting.ID)
+	}
+
+	return watch, nil
+}
+
+func convertLocation(input api.LocationInput) models.Location {
```
Schparky, line 233: I suspect we may need a `convertLocation` for output types.
```suggestion
func convertLocationInput(input api.LocationInput) models.Location {
```

application/actions/watches_test.go
```diff
@@ -55,6 +57,149 @@ func createFixturesForWatches(as *ActionSuite) watchFixtures {
...
+			}
+
+			if tc.watch.MeetingID == "" {
+				as.False(dbWatch.MeetingID.Valid, "expected a null Watch MeetingID")
+			} else {
+				as.Equal(tc.watch.MeetingID, dbWatch.Meeting.UUID.String(), "incorrect Watch Meeting")
+			}
+
+			as.Equal(tc.watch.SearchText, dbWatch.SearchText.String, "incorrect Watch search text")
+			as.Equal(tc.watch.Size.String(), dbWatch.Size.String(), "incorrect Watch search text")
```
Schparky, line 198:
```suggestion
			as.Equal(tc.watch.Size.String(), dbWatch.Size.String(), "incorrect Watch size text")
```

application/models/requestsize.go
```diff
@@ -26,3 +28,14 @@ func (r RequestSize) isLargerOrSame(other RequestSize) bool {
 
 	return sizes[r] <= sizes[other]
 }
+
+func GetRequestSizeFromAPISize(apiSize api.RequestSize) RequestSize {
```
Schparky, line 32: Can `models` just use the type in `api`?

Schparky (approved):

application/actions/watches.go
Baggerone: :+1:

application/models/requestsize.go
Baggerone: I was thinking that since gql is relying on the models.RequestSize, it would be easier to wait until we have removed gql to make that change.

Schparky: yep, good call. I later ran into that myself

PR #347, opened 2021-07-01
Feature/requests list caching

chamberlainpj:
Using https://github.com/go-redis/cache/, I've created a simple caching approach to improve performance of fetching the /requests endpoint to get the list of requests visible to the current user. When a request is created or updated, the respective cache entry (or entries in the case of an update that changes request visibility from same/trusted to all) is updated, and the /requests endpoint composes a list of cached requests (fetching from the database on cache misses).

Key things to review/improve **before** this pull request can be merged:

- Module structure: the _cache_ needs to be able to convert modules.request into api.request, so in the old module structure the cache would have to include _models_, _api_, and _actions_. But _actions_ has to call _cache_ to use the cache in fetching the list of visible requests. My solution was to create a separate _conversions_ module for functions that convert from _model_ to _api_ types, but I'm not sure if this is the best approach.
- Where to update requests on change: currently, I'm doing this from the to-be-replaced graphql code as we haven't yet migrated to a RESTful endpoint for request creation and updates. If it weren't for the module issues, the models might be a nice place to call the cache, but I don't think that will work as the cache needs to call several different _models_ functions.
- Unit testing: after determining a good module structure, I need to fix broken _actions_ unit tests calling _conversion_ functions, and make good unit tests
- Load testing: concurrently to updating this PR based on your feedback, I'll be working to load test these changes and see how/to what extent they improve performance.
- Atomicity? : when a request changes visibility from same to public, it needs to be removed from the organizational cache entry and added to the public cache entry. These are inherently two separate operations, which could cause an issue/temporarily invalid or duplicated request data if one write succeeds and the other fails. But the application would also receive an error, so this may or may not be a concern


application/cache/cache.go
```diff
@@ -0,0 +1,233 @@
...
+
+var (
+	RequestsCache         *cache.Cache
+	PublicRequestsKeyName string = "public"
+)
+
+func init() {
+	ring := redis.NewRing(&redis.RingOptions{
+		Addrs: map[string]string{
+			"redis": "redis:6379",
```
fillup, line 24: These will need to be updated to obtain from the env. Can this package import `domain`? We load most of the env config there and then access them via `domain.WhateverConfig`

application/cache/cache.go
```diff
@@ -0,0 +1,233 @@
+package cache
+
+import (
+	"context"
+	"encoding/json"
+	"time"
+
+	"github.com/go-redis/cache/v8"
```
fillup, line 8: Since our package is named cache and this github.com/go-redis/cache package exports a `cache` package as well we probably need to alias this import. Something like `rediscache "github.com/go-redis/cache/v8"`.  Another option would be to rename our package, but since I think we're wrapping all the redis cache operations with our cache package it might be cleaner just to alias the redis cache package.

application/cache/cache.go
```diff
@@ -0,0 +1,233 @@
...
+		Redis:      ring,
+		LocalCache: cache.NewTinyLFU(1000, time.Minute),
+		// use json in lieu of msgpack for encoding/decoding
+		Marshal:   json.Marshal,
+		Unmarshal: json.Unmarshal,
+	})
+}
+
+// Wrapper for cache write
+func CacheWrite(ctx context.Context, organization string, requestsMap interface{}) error {
```
fillup, line 38: This method name is generic, but the operation details are specific to `requests`. We should either make the method name requests specific as well, or allow the method to be more generic. I assume we'll cache other kinds of things over time, so not having to refactor this later would be nice. Maybe having thin wrapper/helper methods for different types so that we can implement the cache key generation rules there, then call a shared Set method like you have below. For requests I'd expect keys like `requests-orgname-only`, `requests-orgname-affiliated`, `requests-allusers`, etc.

application/cache/cache.go
```diff
@@ -0,0 +1,233 @@
...
+
+	// get public requests
+	_, publicRequestsMap, err := getOrCreateCacheEntryPublic(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	// compose them to get list of all requests visible to the current user
+	visibleRequestsList := make([]api.RequestAbridged, len(publicRequestsMap)+len(privateRequestsMap))
+	for _, privateRequest := range privateRequestsMap {
```
fillup, line 63: I'm pretty sure you can just:

```go
visibleRequestsList := []api.RequestAbridged{}
visibleRequestsList = append(visibleRequestsList, privateRequestsMap...)
visibleRequestsList = append(visibleRequestsList, publicRequestsMap...)
```
If you make the list with the length of the other lists, then append the other lists to it the combined list is twice the length with the first half of the elements just zero value elements.

application/cache/cache.go
chamberlainpj: Yes, I'll move the config to there

application/cache/cache.go
chamberlainpj: Thanks for the tip, this team may make a Gopher of me yet :-)

chamberlainpj: I was indeed having an issue with zero value elements, which initializing as a zero-length fixed. Unfortunately, I can't use the ... syntax as the list and map types don't match up.

Makefile
```diff
@@ -1,4 +1,4 @@
-dev: buffalo adminer migrate
+dev: buffalo redis adminer migrate
```
Schparky, line 1: If you add `redis` to the `depends_on` for the containers that need it (`buffalo` and `test`) you don't need to add anything to `Makefile`. Docker compose does a good job of managing dependencies if they're specified in the `docker-compose.yml` file.

application/actions/requests.go
```diff
@@ -27,195 +22,18 @@ func requestsList(c buffalo.Context) error {
...
-
-	photo, err := hydrateRequestPhoto(ctx, request, tx)
-	if err != nil {
-		return api.RequestAbridged{}, err
-	}
-	output.Photo = &photo
-	
-	return output, nil
-}
+	// TODO evaluate assumption that user can only be logged in as a member of one organization
```
Schparky, line 30: While the statement is true (user can only be logged in as a member of one organization) that doesn't mean you can use only the first organization in the list. You'll need to iterate through all of them, I think.

application/cache/cache.go
```diff
@@ -0,0 +1,233 @@
...
+func CacheRead(ctx context.Context, organization string, requestsMap interface{}) error {
+	return RequestsCache.Get(ctx, organization, requestsMap)
+}
+
+// Rebuild cache after a request is created or updated
+func CacheRebuildOnChangedRequest(ctx context.Context, organization models.Organization, request models.Request) error {
+	// update private cache
+	newEntryCreated, requestsMap, err := getOrCreateCacheEntryPrivate(ctx, organization)
+	if err != nil {
+		return err
```
Schparky, line 84: It is generally preferred to add to the error so that debugging is less painful. Sometimes it adds no value, but usually it does. (e.g. `return errors.New("error in cache rebuild: "+err.Error())`)

application/cache/cache.go
```diff
@@ -0,0 +1,233 @@
...
+	}
+	// may need to update existing cache entry with new data
+	if !newEntryCreated {
+		_, cachedPrivately := requestsMap[request.UUID.String()]
+		// was cached privately, but request visbility has completed. Just need to remove from private cache.
```
Schparky, line 89:
```suggestion
		// was cached privately, but request has been completed. Just need to remove from private cache.
```

application/cache/cache.go
```diff
@@ -0,0 +1,233 @@
...
+		if cachedPrivately && isCompleted(request) {
+			delete(requestsMap, request.UUID.String())
+			return CacheWrite(ctx, organization.Name, requestsMap)
```
Schparky: Regarding my earlier comment about adding to errors, places like this may or may not need additional error detail.

application/cache/cache.go
```diff
@@ -0,0 +1,233 @@
...
+	return visibleRequestsList, nil
+}
+
+// Wrapper for cache read
+func CacheRead(ctx context.Context, organization string, requestsMap interface{}) error {
+	return RequestsCache.Get(ctx, organization, requestsMap)
+}
+
+// Rebuild cache after a request is created or updated
+func CacheRebuildOnChangedRequest(ctx context.Context, organization models.Organization, request models.Request) error {
```
Schparky, line 80: Maybe you have something else in mind, but every call to this function passes the organization as `request.organization` so it seems this function could have only two params.

application/cache/cache.go
```diff
@@ -0,0 +1,233 @@
...
+	// update private cache
+	newEntryCreated, requestsMap, err := getOrCreateCacheEntryPrivate(ctx, organization)
+	if err != nil {
+		return err
+	}
+	// may need to update existing cache entry with new data
+	if !newEntryCreated {
```
Schparky, line 87: The content of this `if` block could be a separate function, maybe `updateRequestPrivateCache`, effectively replacing the comment. Sometimes function names do double-duty as code documentation.

application/cache/cache.go
```diff
@@ -0,0 +1,233 @@
...
+// Gets cache entry for organization, or creates one if none exists
+// Returns true if new cache entry was created
+func getOrCreateCacheEntryPrivate(ctx context.Context, organization models.Organization) (bool, map[string]api.RequestAbridged, error) {
+	var requestsMap map[string]api.RequestAbridged
+	// if a cache value does not exist, we create it
+	if err := CacheRead(ctx, organization.Name, requestsMap); err != nil {
+		tx := models.Tx(ctx)
+
+		var org models.Organization
+		if err := org.FindByUUID(tx, organization.UUID.String()); err != nil {
```
Schparky, line 163: If the `organization` variable has its `id` but just needs to be refetched from the DB, you can use `tx.Reload(&organization)`. Or, maybe you don't need to refresh it at all. Not sure what's needed here.

application/cache/cache.go
```diff
@@ -0,0 +1,233 @@
...
+			requestsMap[requestEntry.ID.String()] = requestEntry
+		}
+
+		return true, requestsMap, CacheWrite(ctx, PublicRequestsKeyName, requestsMap)
+	}
+	return false, requestsMap, nil
+}
+
+// we should cache non-completed public requests publicly
+func isPublic(request models.Request) bool {
```
Schparky, line 220: These three functions seem to belong in `models/request.go` as struct-attached functions on `Request`.

Schparky (commented):
Looks like you've considered and addressed all of the complexity of the logic. No small feat!

Makefile
chamberlainpj: Thanks for the tip!

application/cache/cache.go
chamberlainpj: Good point, simplifying to two params.

application/cache/cache.go
chamberlainpj: I don't actually need to refresh here, as the caller passes in a valid organization. (This was left over from some earlier logic where I was using an api.Request rather than a models.Request and thus didn't have the organization UUID available, and I forgot to take this out.)

application/cache/cache.go
```diff
@@ -0,0 +1,266 @@
...
+		return cacheWrite(ctx, PrivateRequestKeyPrefix+organization.Name, requestsMap)
+	}
+	// was cached privately, but request visibility has changed to public.
+	// Need to remove from private and add to public cache.
+	if cachedPrivately && request.IsPublic() {
+		delete(requestsMap, request.UUID.String())
+		if err := cacheWrite(ctx, PrivateRequestKeyPrefix+organization.Name, requestsMap); err != nil {
+			return err
+		}
+	}
```
Baggerone, line 224: It looks like adding `return nil` would work here and it might make it easier to understand what's going on.
(Same in the next function)

application/cache/cache.go
```diff
@@ -0,0 +1,266 @@
...
+			return errors.New("error in cache rebuild: " + err.Error())
+		}
+	}
+
+	return nil
+}
+
+// Gets cache entry for organization, or creates one if none exists
+// Returns true if new cache entry was created
+func getOrCreateCacheEntryPrivate(ctx context.Context, organization models.Organization) (bool, map[string]api.RequestAbridged, error) {
```
Baggerone, line 151: This function and the next or so similar, I'm tempted to look for a way to factor out some of the code into a third shared function.  But, it's probably not worth it at this point.

Baggerone (approved):

application/cache/cache.go
```diff
@@ -0,0 +1,266 @@
...
+		if !newEntryCreated {
+			if err := updateRequestPrivateCache(ctx, request, requestsMap); err != nil {
+				return errors.New("error in cache rebuild: " + err.Error())
+			}
+		}
+	}
+	return nil
+}
+
+// Rebuild cache after a request is updated
```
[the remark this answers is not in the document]
Schparky: The background behind this is that godoc uses the comment as documentation. It just sounds better when it's a full sentence with the function name as the subject. Unlikely we'll run godoc on this code, so it's more important for the sake of consistency and habit.

application/cache/cache.go
```diff
@@ -0,0 +1,266 @@
...
+// Get all requests visible to user in specified organization from cache, fetching data from database if needed
+func GetVisibleRequests(ctx context.Context, orgs []models.Organization) ([]api.RequestAbridged, error) {
+	// get and de-duplicate private requests for all organizations a user belongs to
+	privateRequestsMap := make(map[string]api.RequestAbridged)
+	for _, organization := range orgs {
+		_, privateRequestsMapPartial, err := getOrCreateCacheEntryPrivate(ctx, organization)
+		if err != nil {
+			return nil, errors.New("error in cache get visible requests: " + err.Error())
+		}
+		for requestID, request := range privateRequestsMapPartial {
```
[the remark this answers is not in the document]
chamberlainpj: Lines 63 - 65 (the inner for loop) _are_ necessary to remove duplicated requests for a user in multiple organizations. For example, if a WBT user requests an item visible to WBT and trusted org SIL, another user belonging to both of these organizations would see this request twice. Re-mapping based on request id ensures each request only appears once in the list.

application/cache/cache.go
chamberlainpj: If we changed the return err to return nil, how would we know when a call to cacheWrite failed?

application/cache/cache.go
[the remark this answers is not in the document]
Baggerone: OK. I see. It didn't register that this was happening inside a loop so that `privateRequestsMapPartial` kept getting changed.

application/cache/cache.go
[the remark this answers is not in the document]
chamberlainpj: Thanks, will update!

application/cache/cache.go
chamberlainpj: At this point, I'll leave refactoring as a TODO comment, but agreed they are similar enough that this is worth exploring later.

chamberlainpj: They do use different SQL queries/call different models functions to populate cache entries, so a refactor will require some thoughtfulness.

application/cache/cache.go
Baggerone: Sorry. I was referring  to the end of the outer `if` block, not to the `return err` that's already there.

In other words, if this outer `if` block is entered, does the code progression still need to continue into the rest of the function following the outer `if` block?  It's not that big of a deal, it's just that mentally for me it's easier to see that OK now we can forget the case where cachedPrivately && request.IsPublic.

Don't bother going back to change it. I just wanted to explain what I meant.  It's more of a personal preference than an actual issue.

chamberlainpj: Thanks for the clarification. Now that the updateRequest<Type>Cache logic is in two functions instead of one, the code logic could actually be changed to an if .... else if .... else format, which would improve readability, and is probably worthwhile.

PR #350, opened 2021-07-01
GET /requests/{id}

Schparky:
Looks like "omitempty" isn't the solution I expected. I reserve the right to change my mind a few more times on this, haha.

Changed nullable fields to either a pointer or a `nulls` type. This made the /requests test more complicated since `"photo":null` looks very different than `"photo":{"id":"fc428c80-2832-4a28-8f86-c55cc1b089aa"`.  Also removed pointers from some that are NOT nullable.

Also includes some fixes on GET /requests. Using related items on the fixtures is problematic because they don't come back from the fixture generation fully-loaded. So we have to specifically go fetch that data to use as a test comparison value. Also, there was a `Save` call on the wrong request, so that had strange side-effects. Fixing that, we now get one fewer requests in the response because one is completed.

Some of the json tags were camel case, which is non-standard for our REST APIs. I changed them to snake case, but that means we'll have to change the UI accordingly.

Just realized that I never went back to make a test for `GET /requests/{id}`. I'll do that right after this is merged. Don't want to lose all this typing. Whew.

### DIFF
--- a/application/actions/requests.go
+++ b/application/actions/requests.go
@@ -207,7 +207,11 @@ func convertRequestAbridged(ctx context.Context, request models.Request) (api.Re
 }
 
 func loadRequestCreatedBy(ctx context.Context, request models.Request) (api.User, error) {
-	createdBy, _ := request.GetCreator(models.Tx(ctx))
+	createdBy, err := request.GetCreator(models.Tx(ctx))
+	if err != nil {
+		return api.User{}, errors.New("loading request creator, " + err.Error())
+	}
+
 	outputCreatedBy, err := convertUser(ctx, *createdBy)
 	if err != nil {
 		err = errors.New("error converting request created_by user: " + err.Error())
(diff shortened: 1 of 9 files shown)


application/actions/requests.go
```diff
@@ -185,36 +249,96 @@ func hydrateRequestDestination(ctx context.Context, request models.Request, tx *
...
+func loadRequestOrganization(ctx context.Context, request models.Request) (api.Organization, error) {
+	organization, err := request.GetOrganization(models.Tx(ctx))
+	if err != nil {
+		err = errors.New("error converting request organization: " + err.Error())
+		return api.Organization{}, err
+	}
+
+	var outputOrganization api.Organization
+	if err := api.ConvertToOtherType(organization, &outputOrganization); err != nil {
+		err = errors.New("error converting organization to api.File: " + err.Error())
```
chamberlainpj, line 319:
```suggestion
		err = errors.New("error converting organization to api.Organization: " + err.Error())
```

chamberlainpj (approved):
Good improvements to GET /requests unit testing and better standardization :-)

Nice job on the /request/{id} handler as well.

Baggerone (approved):

PR #354, opened 2021-07-07
test GET /requests/{id}

Schparky:
Implemented a test for the GET /requests/{id} endpoint

In the process, found a couple issues:
- the authz check responded with a generic 500 because no error message was assigned
- the not-found condition produced a 500 rather than a 404

### DIFF
--- a/application/actions/requests_fixtures_test.go
+++ b/application/actions/requests_fixtures_test.go
@@ -47,7 +47,7 @@ func createFixturesForRequestQuery(as *ActionSuite) RequestQueryFixtures {
 	org2 := models.Organization{Name: "org2", AuthType: AuthTypeGoogle, AuthConfig: "{}"}
 	as.NoError(org2.Save(as.DB))
 
-	requests := test.CreateRequestFixtures(as.DB, 4, true)
+	requests := test.CreateRequestFixtures(as.DB, 4, true, users[0].ID)
 	requests[0].Status = models.RequestStatusAccepted
 	requests[0].ProviderID = nulls.NewInt(users[1].ID)
 
--- a/application/actions/requests_fixtures_test.go
+++ b/application/actions/requests_fixtures_test.go
@@ -44,14 +44,19 @@ func createFixturesForRequestQuery(as *ActionSuite) RequestQueryFixtures {
 	org := userFixtures.Organization
 	users := userFixtures.Users
 
-	requests := test.CreateRequestFixtures(as.DB, 3, true)
+	org2 := models.Organization{Name: "org2", AuthType: AuthTypeGoogle, AuthConfig: "{}"}
+	as.NoError(org2.Save(as.DB))
+
+	requests := test.CreateRequestFixtures(as.DB, 4, true)
 	requests[0].Status = models.RequestStatusAccepted
 	requests[0].ProviderID = nulls.NewInt(users[1].ID)
-	as.NoError(as.DB.Save(&requests))
 
 	requests[2].Status = models.RequestStatusCompleted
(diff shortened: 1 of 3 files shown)


application/actions/requests_test.go
```diff
@@ -840,6 +841,63 @@ func (as *ActionSuite) Test_RequestActions() {
...
+			name:       "bad request ID",
+			user:       f.Users[1],
+			requestID:  "1",
+			wantStatus: http.StatusBadRequest,
+		},
+		{
+			name:       "not found",
+			user:       f.Users[1],
+			requestID:  domain.GetUUID().String(),
+			wantStatus: http.StatusNotFound,
```
chamberlainpj, line 875: What's the difference between a bad request ID and a request ID that's not found? Is the idea that the bad request ID is of an invalid format/type mismatch to the UUID, and the not found is a UUID that isn't a request ID on record?

chamberlainpj (approved):

application/actions/requests_test.go
```diff
@@ -840,6 +841,63 @@ func (as *ActionSuite) Test_RequestActions() {
...
+			req := as.JSON("/requests/" + tt.requestID)
+			req.Headers["Authorization"] = fmt.Sprintf("Bearer %s", tt.user.Nickname)
+			req.Headers["content-type"] = "application/json"
+			res := req.Get()
+
+			body := res.Body.String()
+			as.Equal(tt.wantStatus, res.Code, "incorrect status code returned, body: %s", body)
+
+			if tt.wantStatus == http.StatusOK {
+				as.Contains(body, fmt.Sprintf(`"id":"%s"`, tt.requestID))
```
Baggerone, line 895: Maybe add some more Contains with various fields, including related objects?
(If so, maybe make the conditional the opposite of what it is with a `return` in it)

application/actions/requests_fixtures_test.go
```diff
@@ -44,14 +44,19 @@ func createFixturesForRequestQuery(as *ActionSuite) RequestQueryFixtures {
 	org := userFixtures.Organization
 	users := userFixtures.Users
 
-	requests := test.CreateRequestFixtures(as.DB, 3, true)
+	org2 := models.Organization{Name: "org2", AuthType: AuthTypeGoogle, AuthConfig: "{}"}
+	as.NoError(org2.Save(as.DB))
+
+	requests := test.CreateRequestFixtures(as.DB, 4, true)
```
Baggerone, line 50: Do you want to give it a specific UserID to avoid weird db fetching behavior?

Baggerone (approved):
A couple of suggestions

application/actions/requests_test.go
Schparky: That's correct

application/actions/requests_test.go
Schparky: I can definitely do that. My rationale, though, was that this is specifically testing the handler. The models tests and the `convertRequest` test should verify the actual content.

application/actions/requests_fixtures_test.go
Schparky: Good idea

application/actions/requests_test.go
Baggerone: I am thinking that it would be possible that a future change to the code could sabotage the loading of the related objects in the Action and that none of the other tests would catch that.  (If they do, then that's fine.)

Schparky: I think the other tests stand a better chance of catching that. That is, unless the entire test is duplicated here.

PR #355, opened 2021-07-07
Get events (meetings)

Baggerone:


application/actions/meetings.go
```diff
@@ -1,11 +1,183 @@
...
+		}
+	}
+
+	return output, nil
+}
+
+// converts a model.Meeting into api.Meeting
+func convertMeeting(ctx context.Context, meeting models.Meeting, user models.User) (api.Meeting, error) {
+	output := convertMeetingAbridged(meeting)
+
```
Schparky, line 57: My last refactor on `convertRequest` did one Load up front and then mostly just type conversions after that. You could try that here, but I'm not convinced it is better. The one tangible benefit was that some of the helper functions didn't have an error return.

application/actions/meetings.go
```diff
@@ -1,11 +1,183 @@
...
+		err = errors.New("error converting meeting image file: " + err.Error())
+		return nil, err
+	}
+
+	if imageFile == nil {
+		return nil, nil
+	}
+
+	var outputImage api.File
+	if err := api.ConvertToOtherType(imageFile, &outputImage); err != nil {
```
Schparky, line 111: Seems like a `convertFile` might be helpful here (and elsewhere)

application/actions/meetings.go
```diff
@@ -1,11 +1,183 @@
...
+			return []api.Meeting{}, err
+		}
+	}
+
+	return output, nil
+}
+
+// converts a model.Meeting into api.Meeting
+func convertMeeting(ctx context.Context, meeting models.Meeting, user models.User) (api.Meeting, error) {
+	output := convertMeetingAbridged(meeting)
```
Schparky, line 56: Nice code reuse idea. I wonder if this can be done with `convertRequest` and `convertRequestAbridged`

Schparky (approved):
Looks good

application/actions/meetings.go
chamberlainpj: +1

application/actions/meetings.go
chamberlainpj: I couldn't get it to work because of type mismatch issues (api.Request and api.RequestAbridged are distinct types, whereas api.Meeting is just one type with null fields for API responses). If we could somehow get rid of the need for the api.RequestAbridged type, then this would indeed reuse lots of code

application/actions/meetings_test.go
```diff
@@ -15,3 +17,43 @@ func (as *ActionSuite) verifyMeeting(expected models.Meeting, actual api.Meeting
...
+	for i := 2; i < 4; i++ {
+		lctn := lctns[i]
+		wantContains = append(wantContains, fmt.Sprintf(`"id":"%s"`, mtgs[i].UUID.String()))
+		wantContains = append(wantContains, fmt.Sprintf(`"name":"%s"`, mtgs[i].Name))
+		wantContains = append(wantContains, fmt.Sprintf(`"start_date":"%s`, mtgs[i].StartDate.Format(domain.DateFormat)))
+		wantContains = append(wantContains, fmt.Sprintf(`"end_date":"%s`, mtgs[i].EndDate.Format(domain.DateFormat)))
+		wantContains = append(wantContains, fmt.Sprintf(`"location":{"description":"%s"`, lctn.Description))
+		wantContains = append(wantContains, fmt.Sprintf(`"country":"%s"`, lctn.Country))
+		wantContains = append(wantContains, fmt.Sprintf(`"latitude":%v`, int(lctn.Latitude.Float64)))
+		wantContains = append(wantContains, fmt.Sprintf(`"longitude":%v`, int(lctn.Longitude.Float64)))
```
chamberlainpj, line 49: If you'd like to, you could do something like the following:

```
wantContains = append(wantContains, []string{
			fmt.Sprintf(`"id":"%s"`, requests[i].UUID),
			fmt.Sprintf(`"nickname":"%s"`, creator.Nickname),
			providerJson,
			fmt.Sprintf(`"title":"%s"`, requests[i].Title),
			fmt.Sprintf(`"country":"%s"`, requests[i].Destination.Country),
			fmt.Sprintf(`"description":"%s"`, requests[i].Origin.Description),
			fmt.Sprintf(`"size":"%s"`, requests[i].Size),
			photoJson,
		}...)
```
which I personally find more readable and nicer-looking

chamberlainpj (approved):

application/actions/meetings.go
Schparky: Ah, good point. If the cache hits the bullseye, we might be able to get rid of it!

application/actions/meetings_test.go
Baggerone: :+1:

PR #356, opened 2021-07-08
Cleanup events based on previous PR

Baggerone:


application/models/meeting.go
```diff
@@ -422,6 +418,17 @@ func convertMeeting(ctx context.Context, meeting Meeting, user User) (api.Meetin
 	return output, nil
 }
 
+// convertMeetingImageFile converts a model.Meeting.ImgFile into an api.File
+// This assumes that the meeting's related ImageFile has already been loaded
```
chamberlainpj, line 422: Is there any way we can enforce that this function is only called when the meeting.ImgFile has already been loaded, or it is just important for the calling code to abide by?

chamberlainpj (approved):

application/models/meeting.go
Baggerone: I think it's OK this way, since it's not an exported function and is only called in one place.  The parallel convert functions in request are like this.

@Schparky what do you think?

Schparky: I think it's OK as it is, for the reasons you give. The truly "safe" way to do it is as it was, but I like this better. Another option would be to add a bool to the `Meeting` that indicates whether the field has been loaded. That's sort of the approach that PHP frameworks use, but it seems unnecessarily complex.

Schparky (approved):

PR #357, opened 2021-07-08
Join an event - creates a MeetingParticipant

Baggerone:


application/actions/meetings_test.go
```diff
@@ -61,3 +62,76 @@ func (as *ActionSuite) Test_MeetingsList() {
...
+	type testCase struct {
+		name       string
+		inviteCode string
+		location   models.Location
+		meeting    models.Meeting
+		user       models.User
+		wantInvite models.MeetingInvite
+	}
+
+	testCases := []testCase{
```
Schparky, line 78: The three test cases all follow the same path in `meetingsJoin` so they're more about testing the logic in the models functions. A test case for a bad/missing meeting might be important, for example. I'm OK with leaving this as it is, though, to get to the point where we can finish the REST conversion sooner.

application/actions/meetings_test.go
```diff
@@ -61,3 +62,76 @@ func (as *ActionSuite) Test_MeetingsList() {
...
+			req := as.JSON("/events")
+			req.Headers["Authorization"] = fmt.Sprintf("Bearer %s", tc.user.Nickname)
+			req.Headers["content-type"] = "application/json"
+			res := req.Post(reqBody)
+
+			body := res.Body.String()
+			as.Equal(200, res.Code, "incorrect status code returned, body: %s", body)
+
+			wantContains := []string{
+				fmt.Sprintf(`"nickname":"%s"`, tc.user.Nickname),
```
Schparky, line 120:
```suggestion
				fmt.Sprintf(`"created_by":{"id":"%s"`, tc.user.UUID),
```
to make sure it's actually checking the creator and not just any participant

application/actions/requests.go
```diff
@@ -39,7 +39,7 @@ func requestsList(c buffalo.Context) error {
 	return c.Render(200, render.JSON(requestsList))
 }
 
-// swagger:operation GET /requests/{request_id} RequestsGet
+// swagger:operation GET /requests/{request_id} Requests GetRequest
```
Schparky, line 42: Good catch

Schparky (approved):

application/actions/meetings_test.go
Baggerone: :+1:

application/actions/meetings_test.go
Baggerone: :+1:

application/actions/requests.go
chamberlainpj: +1

chamberlainpj (approved):

PR #358, opened 2021-07-08
RequestsCreate endpoint (POST /requests)

Schparky:


application/actions/requests.go
```diff
@@ -80,3 +83,107 @@ func requestsGet(c buffalo.Context) error {
...
+		request.OrganizationID = org.ID
+	}
+
+	if input.PhotoID.Valid {
+		if _, err := request.AttachPhoto(tx, input.PhotoID.UUID.String()); err != nil {
+			err = errors.New("file ID not found, " + err.Error())
+			appErr := api.NewAppError(err, api.ErrorCreateRequestPhotoIDNotFound, api.CategoryUser)
+			if domain.IsOtherThanNoRows(err) {
+				appErr.Category = api.CategoryDatabase
+				return request, appErr
```
Baggerone, line 168: It looks like this line might be redundant.

Baggerone (approved):

application/actions/actions_test.go
```diff
@@ -95,3 +99,13 @@ func createFixture(as *ActionSuite, f interface{}) {
 		as.T().FailNow()
 	}
 }
+
+func (as *ActionSuite) verifyResponseData(wantData []string, body string) {
+	var b bytes.Buffer
+	as.NoError(json.Indent(&b, []byte(body), "", "    "))
+	for _, w := range wantData {
+		if !strings.Contains(body, w) {
+			as.Fail(fmt.Sprintf("response data is not correct\nwanted: %s\nin body:\n%s\n", w, b.String()))
```
chamberlainpj, line 108: This error message seems like it would make debugging a test easier, and messages like this are something we may want to start adding to other unit tests using the Contains/string-matching approach.

application/actions/requests_test.go
```diff
@@ -965,6 +965,67 @@ func (as *ActionSuite) Test_requestsGet() {
...
+
+	destination := api.Location{Description: "location description", Country: "XX"}
+
+	goodRequest := api.RequestCreateInput{
+		Destination:    destination,
+		OrganizationID: f.Organization.UUID,
+		Size:           api.RequestSize(models.RequestSizeSmall),
+		Title:          "request title",
+	}
+	badRequest := api.RequestCreateInput{}
```
chamberlainpj, line 979: In addition to the bad and good requests, should we also have a unit test with more than the minimal number of fields, e.g., with origin, photoid, etc. populated in order to verify the code running that logic?

chamberlainpj (approved):

application/actions/requests_test.go
Schparky: I believe the code that handles the data is tested by `Test_convertCreateRequestInput` and `Test_convertRequest`. Do you see something that isn't tested at all?

chamberlainpj: It is! I made the comment before reading through the full PR, but forgot to remove it after seeing that test

PR #359, opened 2021-07-08
updateRequest (PUT /requests/{request_id})

Schparky:


application/actions/requests.go
```diff
@@ -186,3 +185,125 @@ func convertRequestInput(ctx context.Context, input api.RequestCreateInput) (mod
...
+		appError := api.NewAppError(err, api.ErrorUpdateRequest, api.CategoryNotFound)
+		if domain.IsOtherThanNoRows(err) {
+			appError.Category = api.CategoryInternal
+		}
+		return request, appError
+	}
+
+	if input.PhotoID.Valid {
+		if _, err := request.AttachPhoto(tx, input.PhotoID.UUID.String()); err != nil {
+			err = errors.New("file ID not found, " + err.Error())
```
Baggerone, line 259:
```suggestion
			err = errors.New("request photo file ID not found, " + err.Error())
```

application/actions/requests_test.go
```diff
@@ -1273,3 +1273,197 @@ func (as *ActionSuite) verifyRequestCreateInput(ctx context.Context, input api.R
...
+
+	tests := []struct {
+		name       string
+		user       models.User
+		input      interface{}
+		request    models.Request
+		wantStatus int
+	}{
+		{
+			name:       "authn error",
```
Baggerone, line 1301:
```suggestion
			name:       "authz error",
```

application/actions/requests.go
```diff
@@ -186,3 +185,125 @@ func convertRequestInput(ctx context.Context, input api.RequestCreateInput) (mod
...
+	if err := request.FindByUUID(tx, id); err != nil {
+		appError := api.NewAppError(err, api.ErrorUpdateRequest, api.CategoryNotFound)
+		if domain.IsOtherThanNoRows(err) {
+			appError.Category = api.CategoryInternal
+		}
+		return request, appError
+	}
+
+	if input.PhotoID.Valid {
+		if _, err := request.AttachPhoto(tx, input.PhotoID.UUID.String()); err != nil {
```
Baggerone, line 258: I guess I didn't see this when reviewing the RequestCreate, but anyway ...

I'm wondering if in the long run, these lines that actually touch the database (here and below) should be moved to the model's update function and this function should only do the transfer of values from the input object to the new request.

If you don't agree, leave it as is, or if you want, you can make it a "TODO".

application/actions/requests_test.go
```diff
@@ -1273,3 +1273,197 @@ func (as *ActionSuite) verifyRequestCreateInput(ctx context.Context, input api.R
...
+}
+
+func (as *ActionSuite) verifyRequestUpdateInput(ctx context.Context, input api.RequestUpdateInput, oldRequest, newRequest models.Request) {
+	as.Equal(models.CurrentUser(ctx).ID, newRequest.CreatedByID, "CreatedBy is not correct")
+
+	as.Equal(input.Description, newRequest.Description, "Description is not correct")
+
+	var destination models.Location
+	as.NoError(as.DB.Find(&destination, newRequest.DestinationID))
+	if msg := "Destination is not correct"; input.Destination != nil {
```
Baggerone, line 1413: It took me a minute to figure out why you had a mixed conditional. I'm not sure if it's worth it.  
I guess it gains you a line of code and allows using `msg :=` everywhere, instead of `msg :=` the first time and `msg =` after that.

It's interesting that the scope of `msg` must get down into the `else` block.

Baggerone (approved):

application/actions/requests_test.go
Schparky: Are you sure? Even though the http status is "Unauthorized" (which I am thinking is not totally accurate) I think this is an authn error. Not that it matters much, but I'm scratching my head over this one.

chamberlainpj (approved):
These changes generally look good to me, don't have any specific comments.

application/actions/requests_test.go
Baggerone: I was just looking at the wantStatus. It's not worth any more time on it. So, feel free to leave it as is.

application/actions/requests.go
Schparky: Started to work on it because I think you're right, but it's a significant change

application/actions/requests_test.go
Schparky: OK, I wondered if it would be too strange

PR #361, opened 2021-07-09
update request status endpoint (PUT /requests/{request_id}/status)

Schparky:
- also moved the deletion of potential providers to the `Request` `AfterUpdate` callback


Baggerone (approved):
Approved provided you have a good reason to remove the code that destroys potential providers when a request is completed.

PR #362, opened 2021-07-12
Endpoint for Remove Me as a Potential Provider

Baggerone:


application/actions/requests.go
```diff
@@ -351,6 +351,50 @@ func requestsAddMeAsPotentialProvider(c buffalo.Context) error {
...
+	}
+
+	domain.NewExtra(c, "request", id)
+
+	if err := provider.Destroy(tx); err != nil {
+		appError := api.NewAppError(err, api.ErrorRemoveMeAsPotentialProviderDestroyIt, api.CategoryInternal)
+		return reportError(c, appError)
+	}
+
+	return c.Render(200, render.String(id.String()))
```
Schparky, line 395: I think this can just return a 204 with no data.

Schparky (approved):

fillup (approved):

application/actions/requests.go
Baggerone: :+1:

PR #365, opened 2021-07-12
Add location fields

Baggerone:
Also fixed migration file names so `down` files match their `up` file.


application/models/testutils_test.go
```diff
@@ -213,10 +213,17 @@ func createPotentialProviderFixtures(tx *pop.Connection, nRequests, nProviders i
...
+	cities := []string{"Miami", "Toronto", "Mexico City", "Chiang Mai", "Paris", "Port Moresby"}
 	locations := make(Locations, n)
+
+	/* #nosec */
 	for i := range locations {
+		randInt := rand.Intn(6)
 		locations[i] = Location{
-			Country:     countries[rand.Intn(6)],
+			Country:     countries[randInt],
+			State:       states[randInt],
```
Schparky, line 225: Looks like we'll get locations like "Mexico City, FL, FR" Reminds me of those kids' books with mix-and-match animal parts, haha!

application/migrations/20210712231000_add_fields_to_locations.up.fizz
```diff
@@ -0,0 +1,5 @@
+add_column("locations", "state", "string")
```
Schparky, line 1: What did you think of making the lat/long not nullable? Better done in a separate PR?

Schparky (approved):
Could you fill in some of these fields in the grifts?

application/migrations/20210712231000_add_fields_to_locations.up.fizz
Baggerone: Yeah, separate PR.

application/models/testutils_test.go
Baggerone: I intended to make the city, state and country all correspond with each other.  This does that doesn't it?  (I.e. Miami will only go with Florida and the US)

Schparky: Oh, yeah, sorry. I just saw `randInt` on them and my brain though they were all separately random.

PR #372, opened 2021-07-19
remove GraphQL

Schparky:


application/actions/messages_fixtures_test.go
```diff
@@ -6,15 +6,15 @@ import (
 	"github.com/silinternational/wecarry-api/models"
 )
 
-type messageQueryFixtures struct {
+type messageTestFixtures struct {
```
Baggerone, line 9: Not important, really, but do we need "Test" in the middle of the type name?

application/actions/threads_fixtures_test.go
```diff
@@ -8,15 +8,15 @@ import (
 	"github.com/silinternational/wecarry-api/models"
 )
 
-type threadQueryFixtures struct {
+type threadTestFixtures struct {
```
Baggerone, line 11: Same comment for this (no "Test" in name).

Baggerone (approved):

PR #376, opened 2021-07-20
Allow new request to be associated with a meeting

Baggerone:


Schparky (commented):
I think the models code needs a change. I get "message": "geo: a valid geo coordinate must be given |country: country must be ISO 3166–1 alpha-2 |description: Description can not be blank. |" when I try to create a request with no destination.

Schparky:
> I think the models code needs a change. I get "message": "geo: a valid geo coordinate must be given |country: country must be ISO 3166–1 alpha-2 |description: Description can not be blank. |" when I try to create a request with no destination.

I wonder if we should make `destination` nullable in the API and in the database, or require the UI to send the event/meeting location in as the request destination.

Baggerone:
I don't think a meeting based request should automatically get the destination of the meeting.  There may be times that the meeting is the origin and a different location is the destination. So, maybe make the UI provide a destination.

Schparky:
> I don't think a meeting based request should automatically get the destination of the meeting. There may be times that the meeting is the origin and a different location is the destination. So, maybe make the UI provide a destination.

Oh, that rocks my world. Hadn't ever considered that the meeting might be the origin. We might have to talk that over. Thoughts on this, @fillup ?

Schparky (approved):
It looks good. I'll go ahead and make the UI send a destination and we can work out the event-as-origin thing later.

PR #380, opened 2021-07-22
Request middleware

Baggerone:
Primarily for discussion. 

Did I implement the kind of thing you had in mind?


application/actions/extrasmware.go
```diff
@@ -0,0 +1,16 @@
+package actions
+
+import (
+	"github.com/gobuffalo/buffalo"
+)
+
+func setHTTPExtras(next buffalo.Handler) buffalo.Handler {
+	return func(c buffalo.Context) error {
+		newExtra(c, "request_method", c.Request().Method)
```
Schparky, line 9: I think these are already included in the log, right?

application/actions/actions.go
```diff
@@ -122,10 +129,20 @@ func getExtras(c buffalo.Context) map[string]interface{} {
...
-		newExtra(c, param, s)
 		err := fmt.Errorf("invalid %s provided: '%s'", param, s)
 		return uuid.UUID{}, api.NewAppError(err, api.ErrorMustBeAValidUUID, api.CategoryUser)
 	}
 	return id, nil
 }
+
+type handlerData struct {
+	User     models.User
+	ModelObj interface{}
```
Schparky, line 143: I was imagining that each handler group would have its own typo so they wouldn't need to use `interface{}`. But this works too.

Schparky (approved):
Looks about right.

application/actions/extrasmware.go
Baggerone: I didn't find them in Rollbar when I went looking for them.

application/actions/actions.go
Baggerone: I think you're right.

application/actions/extrasmware.go
Schparky: They're added by `reportError` so if that doesn't get called, because there's no context, for example, they won't be added to extras (probably not sent to rollbar at all). But then again, this middleware won't do any good either, if `reportError` isn't used.

PR #384, opened 2021-07-23
Get a meeting and add authz to meeting update and participants list

Baggerone:


application/actions/meetings.go
```diff
@@ -127,15 +128,21 @@ func meetingsUpdate(c buffalo.Context) error {
 		return reportError(c, err)
 	}
 
+	if !cUser.CanUpdateMeeting(meeting) {
+		err := fmt.Errorf("user %s is not authorized to update meeting %s",
```
Schparky, line 132: Good thinking to add IDs to the error log. Rather than add the user and meeting IDs to the message, it might be better to add them to "extras". That way the rollbar message is identical for all occurrences, so would be grouped better. I think it may recognize the message with IDs as being similar, but I think it's still better to use extras. Also, fwiw, I think the integer ID is fine in an error log. The UUID is mostly for end-user visibility.

application/models/meeting.go
```diff
@@ -405,7 +405,7 @@ func ConvertMeetings(ctx context.Context, meetings []Meeting, user User) ([]api.
 }
 
 // ConvertMeeting converts a model.Meeting into api.Meeting
-func ConvertMeeting(ctx context.Context, meeting Meeting, user User) (api.Meeting, error) {
+func ConvertMeeting(ctx context.Context, meeting Meeting, user User, includeParticipants bool) (api.Meeting, error) {
```
Schparky, line 408: Another technique that makes for more readable (and expandable) code would be to use an options parameter. Something like:
```go
func ConvertMeeting(ctx context.Context, meeting Meeting, user User, options ...string) (api.Meeting, error)
```
called like:
```go
_, _ = ConvertMeeting(ctx, meeting, user, "includeParticipants")
```

application/actions/meetings.go
```diff
@@ -301,3 +308,45 @@ func parseMeetingDates(input api.MeetingInput, modelMtg *models.Meeting) error {
...
+	meeting := models.Meeting{}
+	if err = meeting.FindByUUID(tx, id.String()); err != nil {
+		appError := api.NewAppError(err, api.ErrorMeetingGet, api.CategoryNotFound)
+		if domain.IsOtherThanNoRows(err) {
+			appError.Category = api.CategoryInternal
+		}
+		return reportError(c, appError)
+	}
+
+	canViewParticipants, err := cUser.CanCreateMeetingInvite(tx, meeting)
```
Schparky, line 340: Should this be a separate authz function like `CanViewParticipants`?

Schparky: Or maybe a more broad `CanManageMeeting` that both can use?

application/actions/meetings_test.go
```diff
@@ -336,8 +346,8 @@ func (as *ActionSuite) Test_meetingsJoin() {
 
 			wantContains := []string{
 				fmt.Sprintf(`"created_by":{"id":"%s"`, mtgCreator.UUID),
-				fmt.Sprintf(`"participants":[{"user":{"id":"%s"`, tc.user.UUID.String()),
-				fmt.Sprintf(`"nickname":"%s"`, tc.user.Nickname),
+				fmt.Sprintf(`"participants":[]`),
```
Schparky, line 349: this is good

application/api/errorcodes.go
```diff
@@ -83,11 +83,12 @@ const (
 
 	ErrorMeetingsGet             = ErrorKey("ErrorMeetingsGet")
 	ErrorMeetingsConvert         = ErrorKey("ErrorMeetingsConvert")
-	ErrorCreateMeeting           = ErrorKey("ErrorCreateMeeting")
```
Schparky, line 86: thanks for renaming this

Schparky (approved):
Some optional suggestions for you

application/actions/meetings.go
Baggerone: :+1:

application/actions/meetings.go
Baggerone: :+1:

application/models/meeting.go
fillup: True, or use constants for `IncludeParticipants` and `ExcludeParticipants` so it reads clear and you don't have to remember which is `true` or `false`

application/actions/meetings.go
fillup: I think we've talked in the past about in some events we'd want to let participants invite other participants so it's not only controlled by the host. It was something about visibility of the event and whether an invitation is required to join. So basing whether they can view participants on the same check could become problematic later.

PR #385, opened 2021-07-23
Delete meeting

Baggerone:


fillup (approved):

Schparky (approved):
Unfortunately, basing the delete on whether or not it has requests means that the `GET /events` or the `GET /events/{id}` needs to return either a `canDelete` property or a `requests` list so the UI knows whether to enable the delete button.

Baggerone:
I'll work on adding a calculated field (CanDelete) on the events on Monday.

Schparky:
> I'll work on adding a calculated field (CanDelete) on the events on Monday.

Sounds good. Since it involves a query on the `requests` table, it may be better to put it on the `GET /events/{id}` response. I think that's ok because we intend to add the delete  button on the event detail page.

PR #387, opened 2021-07-26
Feature/meeting can delete

Baggerone:


fillup (approved):

application/api/meetings.go
```diff
@@ -21,6 +21,11 @@ type Meeting struct {
 	// example: 63d5b060-1460-4348-bdf0-ad03c105a8d5
 	ID uuid.UUID `json:"id"`
 
+	// Whether meeting is deletable by current user
+	// Note: This will only be a valid value when a single
+	//       event/meeting is requested (not for a list of events)
+	IsDeletable bool `json:"is_deletable"`
```
Schparky, line 27: Can this use nulls.Bool so it can be omitted when not valid?

Baggerone: It comes through as `"is_deletable":null`.  Is that OK?

Baggerone: Do you want it to be a `*bool`? That way `omitempty` works on it and keeps it out of the http response.

Schparky: Doesn't matter much from the UI perspective. Whichever you think looks better in the backend code.

PR #395, opened 2021-07-29
Delete meeting participant

Baggerone:


application/models/meetingparticipant.go
```diff
@@ -161,3 +174,10 @@ func (m *MeetingParticipant) createWithoutInvite(tx *pop.Connection, user User,
 	}
 	return nil
 }
+
+// TODO consider adding checks such as ensuring there aren't any requests associated
+// TODO   with the related meeting and user
```
fillup, line 179: I agree. I'm thinking if the user does not have any commitments to carry for someone else, AND they do not have any requests with committed providers, then we can delete it. Otherwise we need to think through releasing committed carries before removing the user and their requests.

PR #409, opened 2022-03-22
Tweaks related to Rollbar messages

Baggerone:


Schparky (approved):
Thanks for cleaning this up

PR #412, opened 2022-03-24
Fix origin match panic

Baggerone:
I also switched to use a local.env file, like in cover. So, if you try to run this locally, you'll need to make that switch too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,12 @@ This software needs docker-compose version 1.24.0 at least.
 In an appropriate directory:
 * git clone git@github.com:silinternational/wecarry-api.git
 * cd wecarry-api
-* cp .env.example .env
+* cp local-example.env local.env
+* give the entries in local.env the appropriate values
+
+  (note that for deployment to the cloud, you need to provide 
+    (e.g. via an `.env.encrypted` file) the proper settings based
+     on the `local-example.env` file)
 
 In application/grifts, copy private.go.example to private.go and 
 correct the example values.
--- a/README.md
+++ b/README.md
@@ -16,8 +16,8 @@ In an appropriate directory:
 * cp local-example.env local.env
 * give the entries in local.env the appropriate values
 
-  (note that for deployment to the cloud, you need to provide 
-    (e.g. via an `.env.encrypted` file) the proper settings based
+  (note that for deployment to the cloud, you need to provide, 
+    e.g. via an `.env.encrypted` file, the proper settings based
      on the `local-example.env` file)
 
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -13,7 +13,12 @@ This software needs docker-compose version 1.24.0 at least.
 In an appropriate directory:
 * git clone git@github.com:silinternational/wecarry-api.git
 * cd wecarry-api
-* cp .env.example .env
+* cp local-example.env local.env
```
Schparky, line 16: @forevermatt and I were discussing this a few days ago. If we ever want to add a '*.env' to all '.gitignore' files to protect against accidentally committing secrets, this wouldn't work. I think we usually put the `.example` on the end. I'm not totally sold on the idea, since IDEs use the last extension to determine the file type. So use your best judgment until we establish a standard.

README.md
```diff
@@ -13,7 +13,12 @@ This software needs docker-compose version 1.24.0 at least.
...
+* give the entries in local.env the appropriate values
+
+  (note that for deployment to the cloud, you need to provide 
+    (e.g. via an `.env.encrypted` file) the proper settings based
```
Schparky, line 20:
```suggestion
    in an `.env.encrypted` file the proper settings based
```
to avoid nested parens. You could also use a comma-separate phrase if you prefer that.

README.md
```diff
@@ -13,7 +13,12 @@ This software needs docker-compose version 1.24.0 at least.
...
+     on the `local-example.env` file)
```
Schparky, line 21: If you change line 16, remember to change this.

Schparky (approved):

docker-compose.yml
```diff
@@ -11,7 +11,7 @@ services:
       - "3000:3000"
       - "6060"
     env_file:
-      - .env
+      - local.env
```
Schparky, line 14: Maybe leave the .env here to not break existing local setups?

Baggerone: The problem with using the `.env` here is that the values would be different for local setups than for what gets deployed via codeship.  E.g., we need the UIURL to be wecarry.local:5001 locally, but not in the .env.encrypted file.

Baggerone: Can we just tell everyone to copy their `.env` file to `local.env` ?

Schparky: Yes, I was just thinking of a way to avoid the trouble. When things like this change, it's often weeks later that someone hits a strange problem and doesn't know how to solve it. I guess in this case, the error message would be pretty clear, so it's fine as you have it.

PR #423, opened 2022-09-30
fix Buffalo worker

Schparky:
After Buffalo update, worker wasn't started automatically. Switch to using the worker included in the app object.

### DIFF
--- a/application/job/job.go
+++ b/application/job/job.go
@@ -18,7 +18,7 @@ const (
 	TokenCleanup     = "token_cleanup"
 )
 
-var Worker *worker.Worker
+var w *worker.Worker
 
 var handlers = map[string]func(worker.Args) error{
 	NewThreadMessage: newThreadMessageHandler,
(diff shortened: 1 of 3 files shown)


Baggerone (approved):

application/job/job.go
```diff
@@ -27,10 +27,10 @@ var handlers = map[string]func(worker.Args) error{
 	TokenCleanup:     tokenCleanupHandler,
 }
 
-func init() {
-	w = worker.NewSimple()
+func Init(appWorker *worker.Worker) {
+	w = appWorker
```
jason-jackson, line 31: Could you do this instead and keep w as worker.Worker? Would keep it simple later on, but I could see it maybe throwing errors too.
```
	w = *appWorker
```

Schparky: That's a good idea. I'll give it a shot.

Schparky: It worked in my local environment, but it wasn't enough to put the broken test back in. And I'm not sure about using a copy of the worker. Maybe best to keep it as a pointer to the original.

PR #425, opened 2023-01-12
Email about outdated requests

Baggerone:
In relation to https://trello.com/c/RZpm19OH/542-check-in-with-users-once-their-needed-by-date-is-past

### DIFF
--- a/application/actions/service_test.go
+++ b/application/actions/service_test.go
@@ -57,6 +57,12 @@ func (as *ActionSuite) Test_serviceHandler() {
 			requestBody: `{"task":"token_cleanup"}`,
 			wantTask:    ServiceTaskTokenCleanup,
 		},
+		{
+			name:        "outdated requests",
+			token:       domain.Env.ServiceIntegrationToken,
+			requestBody: `{"task":"outdated_requests"}`,
+			wantTask:    ServiceTaskOutdatedRequests,
+		},
 	}
 	for _, tt := range tests {
 		as.T().Run(tt.name, func(t *testing.T) {
--- a/application/actions/service_test.go
+++ b/application/actions/service_test.go
@@ -14,9 +14,15 @@ import (
 	"github.com/gobuffalo/httptest"
 	"github.com/silinternational/wecarry-api/api"
 	"github.com/silinternational/wecarry-api/domain"
+	"github.com/silinternational/wecarry-api/job"
 )
 
 func (as *ActionSuite) Test_serviceHandler() {
+
+	getPost := func(name string) string {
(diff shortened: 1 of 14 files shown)


application/actions/service_test.go
```diff
@@ -14,9 +14,15 @@ import (
 	"github.com/gobuffalo/httptest"
 	"github.com/silinternational/wecarry-api/api"
 	"github.com/silinternational/wecarry-api/domain"
+	"github.com/silinternational/wecarry-api/job"
 )
 
 func (as *ActionSuite) Test_serviceHandler() {
+
+	getPost := func(name string) string {
```
Schparky, line 22: Maybe a name like `getPostBody` would add a little more clarity. And I guess I'd have put it at package scope (rather than function scope) but that's just a personal preference.

application/job/job_fixtures_test.go
```diff
@@ -22,6 +27,27 @@ func createFixture(js *JobSuite, f interface{}) {
...
+	users := test.CreateUserFixtures(js.DB, 2).Users
+	requests := test.CreateRequestFixtures(js.DB, 2, false, users[0].ID, users[1].ID)
+	db := js.DB
+
+	requests[0].Status = models.RequestStatusAccepted
+	js.NoError(requests[0].Update(db), "error modifying request for test")
+
+	pastDate := nulls.NewTime(time.Now().Add(-1 * domain.DurationDay))
+	requests[1].NeededBefore = pastDate
+	// avoid validation error for value in the past
```
Schparky, line 40: FYI, you can use Pop without triggering validation. Just use `tx.Update` instead of `tx.ValidateAndUpdate`.

application/models/request.go
```diff
@@ -526,6 +526,15 @@ func (r *Request) AfterCreate(tx *pop.Connection) error {
 	return nil
 }
 
+func (r *Requests) FindOpenPastNeededBefore(tx *pop.Connection, eagerFields ...string) error {
+	now := time.Now().UTC()
+	queryStr := "status = ? and needed_before is not null and needed_before < ?"
```
Schparky, line 531: I like to see SQL keywords in all caps. But not a big deal.

Schparky (approved):

application/actions/service_test.go
Baggerone: I'll make it `postBody`, since I think I remember you don't like to see the prefix `get` on function names and since it's only at function scope. :-)

Schparky: You remember correctly. :-) But it's only a slight preference.

PR #430, opened 2023-02-27
Release -- update Go to 1.19 and update dependencies

Schparky:


application/auth/linkedin/provider.go
```diff
@@ -258,7 +262,11 @@ func (p *Provider) FetchUser(session goth.Session) (goth.User, error) {
 	if err != nil {
 		return user, err
 	}
-	defer respEmail.Body.Close()
+	defer func() {
+		if err := resp.Body.Close(); err != nil {
```
Baggerone, line 266:
```suggestion
		if err := respEmail.Body.Close(); err != nil {
```

Baggerone (approved):
One necessary change, I believe.

application/auth/linkedin/provider.go
Schparky: Ohh, thanks for the save

PR #434, opened 2023-04-26
revise logging and include Sentry

briskt:


application/auth/azureadv2/provider.go
```diff
@@ -278,7 +278,7 @@ func (p *Provider) AuthCallback(c buffalo.Context) auth.Response {
 
 	msg := auth.CheckSessionStore()
 	if msg != "" {
-		domain.Logger.Printf("got message from AzureAD's CheckSessionStore() in AuthCallback ... %s", msg)
+		log.Errorf("got message from AzureAD's CheckSessionStore() in AuthCallback ... %s", msg)
```
Baggerone, line 281: Do you not want `.WithContext` in this call? (Just checking.)
I'll leave it to you to look for any other places it should be included, if that's the case.

Baggerone (approved):

application/auth/azureadv2/provider.go
Schparky: I didn't look for places that had a context but we weren't using it before. But yes, it would be good to add WithContext here.

